PR #17, opened 2015-01-13
Use AllowShutdown for KDM (bnc#907907)

ancorgs:


src/include/security/levels.rb
```diff
@@ -115,7 +115,7 @@ def initialize_security_levels(include_target)
             "FAIL_DELAY"                                => "6",
             "GID_MAX"                                   => "60000",
             "GID_MIN"                                   => "1000",
-            "DISPLAYMANAGER_SHUTDOWN"                   => "root",
+            "AllowShutdown"                             => "Root",
```
jreidinger: really root with big `R` what it is? it is not user name

src/scrconf/kde4_kdmrc.scr
```diff
@@ -0,0 +1,23 @@
+/**
+ * File:	kde4_kdmrc.scr
```
jreidinger: this file missing in Makefile.am

src/scrconf/kde4_kdmrc.scr
```diff
@@ -0,0 +1,23 @@
...
+ * Access:	read / write
+ *
+ * Read/Sets the values defined in
+ * <tt>/usr/share/kde4/config/kdm/kdmrc</tt>
+ */
+.kde4.kdmrc
+
+`ag_ini(
+  `IniAgent(
+    "/usr/share/kde4/config/kdm/kdmrc",
```
jreidinger: so we really modify configuration in `/usr/share`? it smells for me.

src/include/security/levels.rb
ancorgs: Yes. From /usr/share/kde4/config/kdm/kdmrc
```
# Who is allowed to shut down the system. This applies both to the
# greeter and to the command sockets.
# "None" - no "Shutdown..." menu entry is shown at all
# "Root" - the root password must be entered to shut down
# "All" - everybody can shut down the machine
# Default is All
```
Actually, I'm quite sure it's case insensitive in the end but according to documentation, it should be "Root".

jreidinger: ok, not much intuitive. Looks like all desktop developers enjoy creating another layers of abstraction with own names.

src/scrconf/kde4_kdmrc.scr
ancorgs: Good catch. Fixed.

src/scrconf/kde4_kdmrc.scr
ancorgs: I also found it smelly, but it was stated in the bug report as the proper place. I will ask some KDE packager just to be sure.

ancorgs: Confirmed by tittiatcoke that it's the expected location.

jreidinger:
LGTM

PR #18, opened 2015-06-12
Remove runlevels and use yml file

ancorgs:


ancorgs:
``<<`` cannot be used in frozen strings, and as long as internationalization comes into play, the chances of having frozen strings are very high.

imobach:
LGTM

PR #19, opened 2015-06-12
Take systemd aliases into account

ancorgs:


package/yast2-security.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Jun 12 16:27:53 UTC 2015 - ancor@suse.com
+
+- When checking services, systemd aliases are now taken into
+  account (so, for example, rsyslog is accounted and syslog).
```
lslezak: Um, does not sound clear to me, maybe `accounted and syslog` should be `accounted as syslog` instead??

And BNC or FATE missing (this will be rejected in SLE12-SP1).

ancorgs: No fate number yet because there is no version bump yet :-) I plan to add the fate when the feature is finished and the version number is increased.

ancorgs: Yes, it should be "AS syslog".

src/include/security/dialogs.rb
```diff
@@ -514,10 +515,12 @@ def OverviewDialog
               Builtins.y2milestone("Client returned %1", client_ret)
 
               if client_ret == :next || client_ret == :ok ||
-                  client_ret == :finish
+                  client_ret == :finish || client_ret == true
                 # update the current value
-                if Builtins.haskey(@link_update_mapping, ret)
-                  Builtins.eval(Ops.get(@link_update_mapping, ret))
+                if @link_update_mapping.has_key?(ret)
+                  Popup.ShowFeedback(_("Analyzing system"), Message.takes_a_while)
```
lslezak: Use the [new `Popup.Feedback`](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Popup.rb#L1078) with block instead (auto closes the popup).

src/modules/Security.rb
```diff
@@ -858,6 +830,48 @@ def Overview
...
+    protected
+
+    # Sets @missing_mandatory_services honoring the systemd aliases
+    def read_missing_mandatory_services
+      log.info("Checking mandatory services")
+
+      @missing_mandatory_services = @mandatory_services.select do |services|
+        enabled = services.any? { |service| Service.enabled?(service) }
+        log.info("Mandatory services #{services} are enabled: #{enabled}")
+        !enabled
```
lslezak: Rather use `reject` and revert the condition to `enabled` result.

src/modules/Security.rb
```diff
@@ -858,6 +830,48 @@ def Overview
...
+
+      enabled_services = SystemdService.all(names: "Names").select(&:enabled?)
+      mandatory = @mandatory_services.flatten
+      @extra_services = enabled_services.select do |service|
+        # A service is 'extra' if it's neither mandatory or optional
+        extra = !mandatory.include?(service.name) && !@optional_services.include?(service.name)
+        # If we failed to find it in the lists by name, try by alias
+        if extra
+          names = service.properties.names
+          if names
```
lslezak: NP: This looks pretty complicated, CodeClimate would definitely complain here, maybe split this into more methods...

test/security_test.rb
```diff
@@ -64,6 +79,15 @@ def enabled?; true; end
         end
       end
 
+      context "with services that are aliases of optional services" do
+        let(:service_names) { %w(ntp rsyslog auditd random kbd anacron postfix) }
```
lslezak: Avoid using `%w`, see https://github.com/SUSE/style-guides/blob/master/Ruby.md#arrays.

ancorgs: If you look at the whole file, ```%w``` is already used all along the test for defining the contexts. Probably it was considered acceptable for that particular use in the previous reviews.

lslezak:
LGTM

PR #20, opened 2015-06-15
Updated list of services according to SUSE's security team

ancorgs:


imobach:
If Travis is happy, I am. LGTM.

PR #21, opened 2015-06-16
Move security levels settings to yaml files

ancorgs:


src/data/security/level1.yml
```diff
@@ -0,0 +1,39 @@
...
+SYS_GID_MAX:                      '499'
+SYS_GID_MIN:                      '100'
+USERADD_CMD:                      "/usr/sbin/useradd.local"
+USERDEL_PRECMD:                   "/usr/sbin/userdel-pre.local"
+USERDEL_POSTCMD:                  "/usr/sbin/userdel-post.local"
+PASSWD_REMEMBER_HISTORY:          '0'
+SYSTOHC:                          'yes'
+SYSLOG_ON_NO_ERROR:               'yes'
+DISPLAYMANAGER_ROOT_LOGIN_REMOTE: 'no'
+DISPLAYMANAGER_XSERVER_TCP_PORT_6000_OPEN: 'no'
```
jreidinger: broken indentation

src/include/security/levels.rb
```diff
@@ -65,127 +68,15 @@ def initialize_security_levels(include_target)
...
-            "SMTPD_LISTEN_REMOTE"                       => "no",
-            "DISABLE_STOP_ON_REMOVAL"                   => "no",
-            "DISABLE_RESTART_ON_UPDATE"                 => "no",
-            "net.ipv4.tcp_syncookies"                   => "1",
-            "net.ipv4.ip_forward"                       => "0",
-            "net.ipv6.conf.all.forwarding"              => "0"
-          }
-        } 
+      @Levels = {}
+      @LevelsNames.keys.each do |level|
```
jreidinger: you can write it as
```
@levels = @LevelsNames.keys.each_with_object({}) do |level, result|
```

src/include/security/levels.rb
```diff
@@ -65,127 +68,15 @@ def initialize_security_levels(include_target)
...
+        lfile = Directory.find_data_file("security/#{level.downcase}.yml")
+        if lfile
+          @Levels[level] = YAML.load_file(lfile) rescue {}
```
jreidinger: this is really not nice as it hide possible errors in file. Consider that you modify file for testing and use wrong syntax and result is that whole file is empty? At least some logging would be nice.

src/include/security/levels.rb
```diff
@@ -65,127 +68,15 @@ def initialize_security_levels(include_target)
...
+        else
+          @Levels[level] = {}
```
jreidinger: no log? I expect at least warning and better error if file not found.

jreidinger: And ideally exception as if I get it, someone choose level and we do not have proper data file for it.

test/levels_test.rb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env rspec
+
+require_relative 'test_helper'
+
+module Yast
+  class LevelsTester
+    include Yast::I18n
```
jreidinger: simple `include I18n` is enough as you are already in yast namespace

jreidinger: in general better to have full client here that include more, so even if security/levels.rb use some shortcuts it will be there so clas `LevelsTester < Client`

src/data/security/level1.yml
ancorgs: I found more annoying the huge distance from the keys to the values that would be introduced by strict indentation than the fact of having just one line (with an insanely long key) breaking the indentation.

jreidinger: OK, fair enough, then I have second question. How it is sorted? I see now obvious key

ancorgs: I simply respected the sorting of the original .rb file, I assumed there was some human logic on it. But I can sort it alphabetically if we prefer so.

test/levels_test.rb
ancorgs: I prefer to use ```Yast::``` in include because it makes everything more obvious to newcomers (and I see no drawback on it, it's actually even more robust).

src/data/security/level1.yml
jreidinger: To be honest I do not know original one, but if I imagine myself that I add new option, I will be very lost where I should place it, so any logic is good, but if it is not obvious one like alphabetic, then please document it.

test/levels_test.rb
jreidinger: OK, no big issue.

jreidinger:
LGTM, just make travis happy ( problem is old testsuite :(

ancorgs:
And that old testsuite will break again as soon as we change the definition of the levels, so I'll convert it to RSpec to make it less painful to update

jreidinger:
@ancorgs yep, that is preferred solution :)

ancorgs:
@jreidinger To be honest, I'm not sure if the old ```LevelX.rb``` test are worth to be kept (rewritten in this case).  ```Security.Write``` is already tested in ```security_test.rb```. Reading of the levels is already tested in ```level_test.rb```.

Other than that, those tests are only useful to ensure that the levels remain unchanged over time... something I'm not sure is actually desirable for a test.

jreidinger:
@ancorgs it is up to you. It is case by case to decide if functionality is already covered, especially corner cases tested in old testsuite.

ancorgs:
@jreidinger I finally added a test (testing the usage of ```Levels``` to set the values of ```Security```) to make sure that nothing is lost by deleting the old testsuite.

ancorgs:
Anyways, Travis is still not happy. Looks like I have to replace even more old tests :(

ancorgs:
@jreidinger not sure if it worth the effort trying to make that part robust. Is already weak by design (as it was the old testsuite where it comes from).

ancorgs:
I fixed it, anyway.

src/include/security/levels.rb
```diff
@@ -65,127 +68,11 @@ def initialize_security_levels(include_target)
...
-            "DISABLE_STOP_ON_REMOVAL"                   => "no",
-            "DISABLE_RESTART_ON_UPDATE"                 => "no",
-            "net.ipv4.tcp_syncookies"                   => "1",
-            "net.ipv4.ip_forward"                       => "0",
-            "net.ipv6.conf.all.forwarding"              => "0"
-          }
-        } 
+      @Levels = @LevelsNames.keys.each_with_object({}) do |level, levels|
+        lfile = Directory.find_data_file("security/#{level.downcase}.yml")
+        raise(Errno::ENOENT, "#{level.downcase}.yml file not found") unless lfile
```
lslezak: Is this extra error handling needed? YAML raises the exception itself:

```
irb(main):001:0> require "yaml"
=> true
irb(main):002:0> YAML.load_file "missingfile"
Errno::ENOENT: No such file or directory @ rb_sysopen - missingfile
```

Besides a very small difference in the exception message it's the same...

lslezak:
(A bit out of scope, but after removing the old testsuite you can quite simply drop Autotools and switch to Rake. See e.g. https://github.com/yast/yast-proxy/commit/e680ec3b0c0fe5dc25f434bad150655de530a694 .)

src/include/security/levels.rb
ancorgs: Not exactly the same. ```Directory.find_data_file``` looks for a file in several locations, returning nil if it's not found.
https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Directory.rb#L125

If we simply ignore the fact that it returned nil and we call ```YAML.load_file(nil)```, the message will be ```TypeError: no implicit conversion of nil into String```. Way less informative.

jreidinger: yep, I agree here with @ancorgs that specific exception message is much better

jreidinger:
LGTM

PR #24, opened 2015-08-19
Backport systemd-enabled version to SLE12-GA

ancorgs:
This backports the version in SLE-12-GA from 3.1.4 to 3.1.11 (see changelog for all the systemd-related improvements).

The only problem I see is related to i18n, there are no translations for the new strings in SLE-12-GA. I don't know whether it's ok.

Tested in a SLE-12-GA virtual machine. Everything seems to work smoothly.


package/yast2-security.spec
```diff
@@ -27,13 +27,16 @@ Group:          System/YaST
 License:        GPL-2.0
 BuildRequires:	doxygen pkg-config perl-XML-Writer update-desktop-files yast2-pam yast2-testsuite
 BuildRequires:  yast2-devtools >= 3.1.10
+# Directory.find_data_file
+BuildRequires:	yast2 >= yast2-3.1.131
```
jreidinger: is it available on SLE-12-GA?

src/modules/Security.rb
```diff
@@ -597,12 +520,136 @@ def WriteConsoleShutdown(ca)
...
 
+    # Write the settings from @Locations to the corresponding files
+    def write_to_locations
+      commitlist = []
+      # NOTE: the call to #sort is only needed to satisfy the old testsuite
+      @Locations.sort.each do |file, vars|
+        vars.each do |var|
+          val = @Settings[var]
+          if val && val != SCR.Read(path("#{file}.#{var}"))
+            SCR.Write(path("#{file}.#{var}"), val)
```
jreidinger: that not nice way to construct path, better is `path(file) + var` see https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/path.rb#L24

ancorgs: Well, this is only backporting. It's the same code in the master branch, so I'd rather leave it as it is. I only want to introduce changes that are needed for the whole thing to run in SLE12-GA.

jreidinger:
LGTM

PR #25, opened 2015-08-20
Merge of #24 back to master

ancorgs:
A lot of repeated commit (result from cherry-picks) but not changes. Just to keep git branches in good sync


jreidinger:
LGTM

PR #29, opened 2015-09-24
Bump version due to SLE branching

ancorgs:


jreidinger:
in general LGTM...but how we handle SLE12? does we need always separate branch? control file based behavior is not possible?

ancorgs:
Well, a full redesign of yast2-security is in the TODO list for both security and yast teams. We didn't find the time in the last CSM Workshop (although we tried). Expect it for SP2 or SLE13. Meanwhile I wouldn't put much effort into the topic.

PR #31, opened 2016-02-26
Remove interpretation of Ctrl + Alt + Del for s390 from Boot Permisions fate#319711

teclator:
I had tested mocking Arch.s390 and also in SP1 but, modifying the code because the AllowShutdown option was not called so and also was dropped from the menu. 

I'm trying to reproduce in a SP2, but if it is not mandatory could be merged.

### DIFF
--- a/src/include/security/dialogs.rb
+++ b/src/include/security/dialogs.rb
@@ -573,7 +573,7 @@ module Yast
     def BootDialog
       # Boot dialog caption
       caption = _("Boot Settings")
-      help = Ops.get_string(@HELPS, "boot", "")
+      help = @HELPS["boot"] || ""
 
       # Boot dialog contents
       contents = HVCenter(
@@ -587,15 +587,7 @@ module Yast
                 _("Boot Permissions"),
                 HBox(
                   HSpacing(3),
-                  VBox(
-                    VSpacing(1),
-                    settings2widget("CONSOLE_SHUTDOWN"),
-                    VSpacing(1.0),
-                    settings2widget("AllowShutdown"),
-                    VSpacing(1.0),
-                    settings2widget("HIBERNATE_SYSTEM"),
-                    VSpacing(1)
-                  ),
+                  boot_vbox_widgets,
                   HSpacing(3)
                 )
               ),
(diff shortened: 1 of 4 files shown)


src/include/security/helps.rb
```diff
@@ -352,8 +331,37 @@ def initialize_security_helps(include_target)
...
+
+    end
+
+    def boot_dialog_help
+      help = _(
+        "<p><b><big>Boot Security</big></b></p>\n<p>In this dialog, change various boot settings related to security.</p>"
+      )
 
+      # Boot dialog help 2/4, not shown for s390 architecture
+      help << _(
```
kobliha: I'm afraid that you'll get an error here. @lslezak ?

src/include/security/helps.rb
```diff
@@ -352,8 +331,37 @@ def initialize_security_helps(include_target)
...
+        "<p><b>Interpretation of Ctrl + Alt + Del</b>:\n" +
+          "Configure what the system should do in response to\n" +
+          "someone at the console pressing the CTRL + ALT + DEL key\n" +
+          "combination. Usually the system reboots. Sometimes it is desirable\n" +
+          "to ignore this event, for example, when the system serves as both\n" +
+          "workstation and server.</p>"
+      ) if !Arch.s390
```
kobliha: Ctrl + Alt + Del is possible on S390, but does not do a reboot - does shutdown instead.

teclator: Yeah, we only are removing the option for redefine it and the help for this option in the corresponding section

src/include/security/dialogs.rb
```diff
@@ -973,5 +965,18 @@ def LoginDialog
 
       deep_copy(ret)
     end
+
+    def boot_vbox_widgets
+      VBox(
+        VSpacing(1),
+        Arch.s390 ? Empty() : settings2widget("CONSOLE_SHUTDOWN"),
```
kobliha: I'm afraid that this is wrong, we want to offer this combobox, but we do not want to have "reboot" there. Just "Ignore" and "Halt".

PR #32, opened 2016-02-29
WIP: First aproach without Overengineering S390 hide booting

teclator:
I will add two approach,
this one
and other moving stuf to a specific class: https://github.com/yast/yast-security/pull/33

### DIFF
--- a/src/include/security/helps.rb
+++ b/src/include/security/helps.rb
@@ -52,28 +52,7 @@ module Yast
             "<p><b><big>Aborting Saving</big></b><br>\nAbort the save procedure by pressing <b>Abort</b>.</p>"
           ),
         # Boot dialog help 1/4
-        "boot"           => _(
-          "<p><b><big>Boot Security</big></b></p>\n<p>In this dialog, change various boot settings related to security.</p>"
-        ) +
-          # Boot dialog help 2/4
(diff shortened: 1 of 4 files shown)


src/include/security/helps.rb
```diff
@@ -352,9 +331,47 @@ def initialize_security_helps(include_target)
...
+    end
+
+    def boot_dialog_help
+      help = _(
+        "<p><b><big>Boot Security</big></b></p>\n<p>In this dialog, change various boot settings related to security.</p>"
+      )
+
+      # Boot dialog help 2/4, not shown for s390 architecture
+      if Arch.s390
+        help += _(
```
mchf: what about ```<<``` instead of ```+=``` ?

teclator: I did so in the closed one, but @kobliha wasn't sure about some kind of problems, so i replaced but maybe he mind other thing in the comment and not about the operator.

mvidner: @mchf We used `<<` in the past but then discovered that if you run in a non-English locale, gettext will give you immutable strings. :bomb:

teclator: @mvidner thnx for the explanation!

PR #33, opened 2016-02-29
Remove reboot entry for s390 from Boot Permisions fate#319711

teclator:
- I moved the behaviour to a module at least the read stuff, write could be moved also.
- Maybe the help message could be redefined to avoid a special case.
- Maybe Inittab for master could be removed, and also could be broken in previous releases as it ever returns nil and it reassign Settings["CONSOLE_SHUTDOWN"] outside see [here](https://github.com/yast/yast-security/blob/master/src/modules/Security.rb#L378)


teclator:
Travis is not happy, i didn't run testsuite tests, so i will check.

package/yast2-security.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Feb 26 12:40:29 UTC 2016 - knut.anderssen@suse.com
+
+- Removed "Boot permissions - Interpretation"  of Ctrl + Alt + Del
+  for s390. (fate#319711)
+- 3.2.1
+
+-------------------------------------------------------------------
```
teclator: This comment should be changed.

lslezak:
Found the reason for the failing old test:

- There are white listed checked files, the SCR calls in the other files are ignored :exclamation:  https://github.com/yast/yast-security/blob/master/testsuite/tests/Read.rb#L13. Simply add `ctrl_alt_del_config.rb` to the white list.
- After this change it still fails. It complains about extra error log produced by this line: https://github.com/yast/yast-security/pull/33/files#diff-4f84cafd33940703e8afd2335f9c42a0R75. The problem is that the `symlink` call returns the default mock value `nil` and this code reports error. Add a mock for `.target.symlink` here: https://github.com/yast/yast-security/blob/master/testsuite/tests/Read.rb#L77

BTW it took me over 1 hour to find this out... :rage: we should really get rid of the old tests :boom:

teclator:
@lslezak thnx for your help and effort!.

The second point is more understandable and i realized it also checking the first point. But the first point is very tricky. 

You are right, i know i have spent more time than spected in this PBI but at this point maybe could be nice to remove old testsuite at least for #Read

src/include/security/helps.rb
```diff
@@ -352,9 +331,48 @@ def initialize_security_helps(include_target)
...
+
+    def boot_dialog_help
+      help = _(
+        "<p><b><big>Boot Security</big></b></p>\n<p>In this dialog, change various boot settings related to security.</p>"
+      )
+
+      # Boot dialog help 2/4, diferent message based on default
+      if ::Security::CtrlAltDelConfig.default == "reboot"
+        help += _(
+          "<p><b>Interpretation of Ctrl + Alt + Del</b>:\n" +
```
lslezak: Use a multiline string here instead of the `+` operator, just replace it by backslash `\`.

src/modules/Security.rb
```diff
@@ -291,46 +293,19 @@ def ReadServiceSettings
       nil
     end
 
+    def inittab_shutdown_configured?
+      inittab = SCR.Dir(path(".etc.inittab"))
+      inittab.include? "ca"
```
lslezak: Missing parenthesis here.

test/security_test.rb
```diff
@@ -207,5 +208,225 @@ def enabled?; true; end
...
       end
     end
+
+    describe "#ReadConsoleShutdown" do
+      let(:ctrl_alt_del_file) { "/etc/systemd/system/ctrl-alt-del.target" }
+      let(:target_link) { "/usr/lib/systemd/system/poweroff.target" }
+
+      context "when systemd is installed" do
+
+        context "in a non s390 architecture" do
```
lslezak: It's **on** an architecture...

test/security_test.rb
```diff
@@ -207,5 +208,225 @@ def enabled?; true; end
...
+              allow(SCR).to receive(:Read).with(path(".target.symlink"), ctrl_alt_del_file)
+                .and_return(target_link)
+
+              expect(Security.ReadConsoleShutdown).to eql("reboot")
+            end
+
+          end
+        end
+
+        context "in a s390 architecture" do
```
lslezak: Same here.

test/security_test.rb
```diff
@@ -207,5 +208,225 @@ def enabled?; true; end
...
+          end
+        end
+      end
+
+      context "when systemd is not installed but inittab exist" do
+        before do
+          allow(Package).to receive(:Installed).with("systemd") { false }
+        end
+
+        it "ever returns nil" do
```
lslezak: **always** returns nil

test/test_helper.rb
```diff
@@ -45,3 +49,5 @@
     ]
   end
 end
+
+# force loading all files to report proper code coverage
```
lslezak: This comment belongs to the added line above?

test/test_helper.rb
```diff
@@ -35,6 +38,7 @@
 
   # for coverage we need to load all ruby files
   Dir["#{SRC_PATH}/modules/**/*.rb"].each { |f| require_relative f }
+  Dir.chdir(libdir) { Dir["**/*.rb"].each { |f| require f } }
```
lslezak: Do we really need `chdir` here? Why is it different from the line above?

src/lib/security/ctrl_alt_del_config.rb
```diff
@@ -0,0 +1,102 @@
...
+        nil
+      end
+
+      def current_systemd
+        if !Yast::FileUtils.Exists(SYSTEMD_FILE)
+          ret = nil
+        else
+          link = Yast::SCR.Read(Yast::Path.new(".target.symlink"), SYSTEMD_FILE).to_s
+          ret = case link
+                when "/usr/lib/systemd/system/poweroff.target"
```
lslezak: Looks like too much indentation here...

imobach: I agree. What about this?

```ruby
ret =
  case link
  when "/usr/lib/systemd/system/poweroff.target"
```

src/modules/Security.rb
teclator: Aren't optional here?

test/security_test.rb
teclator: Fixing my poor english

test/test_helper.rb
teclator: The changes to this file should be reverted.

test/test_helper.rb
teclator: Same as above, the changes to this file are not needed. I will remove it from the PR.

src/lib/security/ctrl_alt_del_config.rb
teclator: Yeah, i was going to write the same alternative as Imo proposed. 

BTW: rubocop is not unhappy just a preferable option to read it.

src/modules/Security.rb
lslezak: Yes, they are optional, but the [style guide](https://github.com/SUSE/style-guides/blob/master/Ruby.md#parameters) allows omitting them only in special cases.

teclator: Hehe, ok, i will put in this case, but in the guide linked for example you have puts 'whatever' that is also a method without parenthesis :P with just one parameter have them or not is not really a problem but i will try to do as proposed in advance

lslezak:
LGTM

src/include/security/helps.rb
```diff
@@ -352,9 +331,48 @@ def initialize_security_helps(include_target)
...
+    end
+
+    def boot_dialog_help
+      help = _(
+        "<p><b><big>Boot Security</big></b></p>\n<p>In this dialog, change various boot settings related to security.</p>"
+      )
+
+      # Boot dialog help 2/4, diferent message based on default
+      if ::Security::CtrlAltDelConfig.default == "reboot"
+        help += _(
```
lslezak: We should probably split the message to the constant part and a variable part to save some work for the translators.

```ruby
if ::Security::CtrlAltDelConfig.default == "reboot"
  # TRANSLATORS: part of help text - default action (the default is reboot)
  details = _("Usually the system reboots...")
else
  # TRANSLATORS: part of help text - default action (the default is halt)
  details = _("By default the system...")
end

# TRANSLATORS: %s is help text - default action
help += _("<p><b>Interpretation of... %s</p>") % details
```

teclator:
* Splited Read in various methods to me more readable/testable. 
* Added rspec tests for the new methods and for Read, in this case just expecting calls for this methods and returning true. At the finish the method does a Settings backup but there is not a method to read the Backup and it hasn't got a publish method, so i left it out.

src/include/security/helps.rb
```diff
@@ -352,9 +331,53 @@ def initialize_security_helps(include_target)
...
+        )
+      end
+
+      # Boot dialog help 2/4
+      # TRANSLATORS: %s is help text - default action
+      help += _(
+        "<p><b>Interpretation of Ctrl + Alt + Del</b>:\n" \
+          "Configure what the system should do in response to\n" \
+          "someone at the console pressing the CTRL + ALT + DEL key\n" \
+          "combination. %s"
```
lslezak: Please move the closing `</p>` tag here from the `details` message. It is difficult to verify that the `<p>`... `</p>` pair is complete. And some translators could be confused and would be tempted to add "missing" tags.

src/modules/Security.rb
```diff
@@ -373,104 +415,29 @@ def Read
 
       # Read security settings
       read_from_locations
-      Builtins.y2milestone("Settings=%1", @Settings)
+      log.info "Settings=#{@Settings}"
```
lslezak: The logging is inconsistent, here is `info` level, the rest is `debug`. At least the final logging should be `info` to log the final value after reading all parts. And also the message is the same, there should be some context to distinguish the steps, e.g.:

```ruby
log.info "Settings (after reading ...): #{@Settings}"
```

src/modules/Security.rb
```diff
@@ -365,6 +340,73 @@ def read_kernel_settings
       end
     end
 
+    def read_encryption_method
```
lslezak: NP: Ideally these methods should be private and only the main `Read` should be public. But there already are other public read methods so let's keep it as it is...

src/include/security/helps.rb
teclator: +1

src/modules/Security.rb
teclator: Exactly, it was the main reason to maintain it public.

src/modules/Security.rb
teclator: +1

src/modules/Security.rb
```diff
@@ -363,6 +340,80 @@ def read_kernel_settings
         val = default_value if val.nil? || val == ""
         @Settings[key] = val
       end
+
+      log.debug "Settings (after #{__callee__}): #{@Settings}"
```
lslezak: Interesting trick... :open_mouth: 

The method names are pretty descriptive in this case, otherwise I'd rather prefer some human readable string than the internal function names.

teclator: The original idea was to define log_read_settings but then __method__ or __callee__ is log_read_settings and not the method_name who is calling this method, i have to research more about introspection but i think it is possible.

The string method is more legible, but if you change the method name you don't have to change the log entry, too lazy?

lslezak: No, my point is that I'd prefer something readable even for non-developers (admins) , e.g. `after reading /etc/sysctl.conf file` is probably be more understandable than the internal function name.

lslezak: But I'm fine with it, in this case the function names are reasonable.

lslezak:
LGTM, let's finally close it :wink:

lslezak:
BTW thanks for converting the old test! :+1:

src/modules/Security.rb
kobliha: BTW, this is another place in Yast where `__callee__` has  been already used: https://github.com/yast/yast-yast2/blob/8eb9214265fdf8afcae1531c5a0309799028d79d/library/packages/src/modules/Product.rb#L49

PR #34, opened 2016-03-07
Added support for different display managers security configuration (bnc#946889)

teclator:
- Testsuite test for Import is failing so i'm cleaning up this method and moving test to Rspec.


src/modules/Security.rb
```diff
@@ -65,8 +74,10 @@ def main
       @optional_services = srv_lists["optional_services"] || []
       # All other services should be turned off
 
+      @display_manager = ::Security::DisplayManager.new
```
teclator: I will change this call with current, as the initial proposal.

src/lib/security/display_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+    attr_reader :name
+
+    def self.current
+      Yast::SCR.Read(Yast::Path.new(CONFIG_PATH))
+    end
+
+    def initialize
+      @name = DisplayManager.current
+    end
+
```
teclator: I will change this an current will be responsable for return an DisplayManager object.

teclator:
At this point, and after check that Write is already covered by rspec tests, we could remove autotools and testsuite tests completely.

package/yast2-security.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Mar  7 09:51:41 UTC 2016 - knut.anderssen@suse.com
+
+- Removing the need for a diferent branch for different display
+  manager behaviour. (bnc#946889).
```
ancorgs: This looks like a message for ourselves, not something user-oriented. The real feature is that it can handle both types of DM

src/lib/security/display_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+    attr_reader :name
+
+    def self.current
+      @display_manager = DisplayManager.new
+    end
+
+    def initialize
+      @name = Yast::SCR.Read(Yast::Path.new(CONFIG_PATH)).to_s
+    end
```
ancorgs: Mmm, I would prefer to have
```ruby
def self.current
  # First of all, no need of  @display_manager
  configured_dm = Yast::SCR.Read(Yast::Path.new(CONFIG_PATH)).to_s
  DisplayManager.new(configured_dm)
end

def initialize(name)
  @name = name
end
```

I think it's more flexible and makes more sense

src/lib/security/display_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def kdm?
+      @name == "kdm"
+    end
+
+    def default_settings
+      { shutdown_key => default_value }
+    end
+
+    def shutdown_key
```
ancorgs: NP (nitpicking): when reading a method `shutdown_key` for a `DisplayManager` class, I mentally connect it with a key in a keyboard to shutdown the system. I don't have a better suggestion, though.

src/lib/security/display_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+    def initialize
+      @name = Yast::SCR.Read(Yast::Path.new(CONFIG_PATH)).to_s
+    end
+
+    def kdm?
+      @name == "kdm"
+    end
+
+    def default_settings
+      { shutdown_key => default_value }
```
ancorgs: Inconsistent naming. What about something like?
```ruby
{ shutdown_setting => shutdown_default_value }
```

src/lib/security/display_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def shutdown_key
+      @shutdown_key ||= kdm? ? "AllowShutdown" : "DISPLAYMANAGER_SHUTDOWN"
+    end
+
+    def default_value
+      @default_value ||= kdm? ? "All" : "all"
+    end
+
+    def options
```
ancorgs: Same inconsistence than above. What about `shutdown_options`?

src/modules/Security.rb
```diff
@@ -293,6 +295,24 @@ def ReadServiceSettings
       nil
     end
 
+    # Delegation methods to @display_manager object
```
ancorgs: What's the problem with exposing `display_manager` directly?

src/modules/Security.rb
```diff
@@ -688,6 +708,11 @@ def Write
       true
     end
 
+    SYSCTL_VALUES = {
```
ancorgs: Better solution than the original, no doubt about it. But I'm not sure if we have some policy about constants defined in the middle of the file. I would have expected it at the top.

test/security_test.rb
```diff
@@ -551,33 +553,64 @@ def enabled?
...
-
       after do
         reset_scr_root
       end
 
-      it "sets login definitions based on /etc/login.defs" do
-        expect(Security.Settings["FAIL_DELAY"]).to eql("3")
-      end
+      context "when display manager is gdm" do
+        before do
```
ancorgs: The usage of `before` is a little bit messy. Let me suggest a better one...

test/security_test.rb
```diff
@@ -551,33 +553,64 @@ def enabled?
...
+          change_scr_root(File.join(DATA_PATH, "system"))
+          allow(SCR).to receive(:Read)
+            .with(path(".sysconfig.displaymanager.DISPLAYMANAGER"))
+            .and_return("gdm")
+          Security.init_settings
+          expect(Security.Settings["DISPLAYMANAGER_SHUTDOWN"]).to eql("all")
```
ancorgs: I never expect an `expect` in a `before`... or I didn't before. :wink:

src/lib/security/display_manager.rb
ancorgs: BTW, what if `.sysconfig.displaymanager.DISPLAYMANAGER` is not set? Shouldn't we have a fallback?

ancorgs: Nevermind, `""` is actually a valid fallback

ancorgs: Well, not that valid, we display the name in the UI in a couple of places. Not sure if `""` is so convenient in those cases... hmmm. Probably I'm thinking to much.

test/security_test.rb
ancorgs:
```ruby
describe "#read_from_locations" do
  after do
    reset_scr_root
  end

  before do
    change_scr_root(File.join(DATA_PATH, "system"))

    allow(SCR).to receive(:Read)
      .with(path(".sysconfig.displaymanager.DISPLAYMANAGER"))
      .and_return(display_manager)
  end

  context "when display manager is gdm" do
    let(:display_manager) { "gdm" }
    
    before do
      Security.init_settings
      Security.read_from_locations
    end

    it "allows everybody to shutdown by default" do
      expect(Security.Settings["DISPLAYMANAGER_SHUTDOWN"]).to eql "all"
    end

    it "bla" do
      ...
    end
  end

  context "when display manager is kdm" do
    # Knut, even if this is the value set in the chroot, I want something more
    # explicit here. It took me a while to guess why you were mocking this value
    # only in the gdm scenario
    let(:display_manager) { "kdm" }
    
    before do
      allow(SCR).to receive(:Read).with(path(".kde4.kdmrc.AllowShutdown"))
        .and_return("All")

      Security.init_settings
      Security.read_from_locations
    end

    it "bla" do
      ...
    end
  end

  context "when display manager is not specified" do
    let(:display_manager) { nil }
    # I want to see this tested also
  end
end
```

src/lib/security/display_manager.rb
teclator: I firstly implemented as you proposed, but then i changed to avoid instantiate it with custom name, so i will propose to do so but privatizing new

About blank names, default for `.sysconfig.displaymanager.DISPLAYMANAGER` is "", but UI is only ready for GDM and KDM.

src/lib/security/display_manager.rb
teclator: shutdown_var_name for example?

src/lib/security/display_manager.rb
teclator: well shutdown_var_name or shutdown_setting ?

src/lib/security/display_manager.rb
teclator: Same as above +1

src/modules/Security.rb
teclator: I was thinking i not expose it, bu there is not a real problem there so +1

src/modules/Security.rb
teclator: Good question, the constants are normally written at top but in Yast i had seen big classes with a lot of lines, and had seen that there is some ones that are written constants declarations just above the method using it.

So in case it will be used in different parts of the code i move it to the top, but in this case you don't have to scroll up to read it. So... what is the preferred way? for me make sense both alternatives

teclator:
In case of blank name for display manager we return nil and we are hiding the combo box from the boot permissions dialog,

ancorgs:
You need to fix .travis, but the code LGTM

PR #36, opened 2016-03-14
Added support for different display managers (bnc#946889)

teclator:


src/include/security/dialogs.rb
```diff
@@ -568,6 +570,21 @@ def OverviewDialog
       deep_copy(ret)
     end
 
+    def vbox_boot_permissions
+      vbox = VBox(VSpacing(1), settings2widget("CONSOLE_SHUTDOWN"))
+
+      if @display_manager
```
lslezak: I'd prefer using `Empty()` widget trick instead of those `vbox <<` additions to be more consistent with the other code. See this [example](https://github.com/yast/yast-registration/blob/7465f820ccb9661828cbb57cb78f83e96c4ef74a/src/lib/registration/ui/import_certificate_dialog.rb#L97).

src/include/security/widgets.rb
```diff
@@ -261,6 +247,22 @@ def initialize_security_widgets(include_target)
             "Value"  => "100"
           }
         }
+
+      @WIDGETS.merge!(
+        @display_manager.shutdown_var_name => shutdown_login_manager_widget
+      ) if @display_manager
```
lslezak: Do not use trailing `if` for multiline command. (We should enable rubocop for this module... :wink:)

src/include/security/widgets.rb
```diff
@@ -261,6 +247,22 @@ def initialize_security_widgets(include_target)
...
+    end
+
+    def shutdown_login_manager_widget
+      {
+        "Widget"  => "ComboBox",
+        # ComboBox label
+        "Label"   => _(
+          "&Shutdown Behaviour of %s Login Manager:"
```
lslezak: Add a `TRANSLATORS:` comment and describe the `%s` meaning (possibly with examples).

src/include/security/widgets.rb
```diff
@@ -271,12 +273,24 @@ def boot_option_labels
       }
     end
 
+    def shutdown_labels
+      {
```
lslezak: NP: This looks like a constant, but with translations it [is a bit tricky](https://github.com/webyast/webyast/wiki/Localization#translating-constants).

src/lib/security/display_manager.rb
```diff
@@ -0,0 +1,81 @@
+# encoding: utf-8
+
+# ------------------------------------------------------------------------------
+# Copyright (c) 2015 SUSE LLC, All Rights Reserved.
```
lslezak: Errr...`2016`?

src/include/security/widgets.rb
teclator: Thnx for this point ;)

src/include/security/widgets.rb
teclator: OK, i will add, it was GDM and KDM before so it should be already present, but it could be other LM.

src/include/security/widgets.rb
teclator: I used it to the code i added and rubocop is not unhappy with this.

src/include/security/widgets.rb
```diff
@@ -261,22 +262,36 @@ def initialize_security_widgets(include_target)
...
+        ) % @display_manager.name,
+        "Options" => shutdown_options,
+        "Value"   => @display_manager.shutdown_default_value
       }
     end
 
     def console_shutdown_options
       ::Security::CtrlAltDelConfig.options.map do |opt|
-        [opt, boot_option_labels[opt]]
+        [opt, BOOT_OPTION_LABELS[opt.downcase]]
```
lslezak: Constants are not translated, you need to translate them explicitly here: `_(BOOT_OPTION_LABELS[opt.downcase])`.

Note: `N_("foo")` does not translate (returns the original), it just marks the text to be found by gettext when creating the pot file. You need to explicitly translate by `_()` when the text is used.

src/include/security/widgets.rb
```diff
@@ -261,22 +262,36 @@ def initialize_security_widgets(include_target)
...
       end
     end
 
-    # EOF
+    def shutdown_options
+      @display_manager.shutdown_options.map do |opt|
+        [opt, SHUTDOWN_LABELS[opt.downcase]]
```
lslezak: Same here.

lslezak:
LGTM :+1:

PR #40, opened 2016-09-01
[WIP] refactoring autoconverted ruby code into ruby code

vpereira:
Hi, I refactored some autoconverted code. I didn't change any functionality. Just refactoring.

### DIFF
--- a/src/clients/security.rb
+++ b/src/clients/security.rb
@@ -30,12 +30,13 @@
 # only some calls to the basic functions. The settings are
 # initialized, main dialog is called and then settings are
 # saved.
+
 module Yast
   class SecurityClient < Client
     def main
       Yast.import "UI"
 
-      #**
+      # **
       # <h3> Security configuration
 
       textdomain "security"
@@ -49,8 +50,6 @@ module Yast
       Yast.import "Security"
 
       Yast.include self, "security/wizards.rb"
-
-
 
       # the command line description map
       @cmdline = {
@@ -89,7 +88,7 @@ module Yast
           }
         },
         "options"    => {
-          "workstation"        => {
+          "workstation" => {
             # command line help text for 'level workstation' option
             "help" => _(
               "Workstation security level"
@@ -142,7 +141,7 @@ module Yast
         "mappings"   => {
           "summary" => [],
           "level"   => ["workstation", "roaming", "server"],
-          #FIXME 1,2,3 aliases
+          # FIXME: 1,2,3 aliases
           "set"     => [
             "passwd",
             "crack",
@@ -158,7 +157,7 @@ module Yast
       # Finish
       Builtins.y2milestone("Security module finished")
       Builtins.y2milestone("----------------------------------------")
-      deep_copy(@ret) 
+      deep_copy(@ret)
 
       # EOF
     end
@@ -180,83 +179,66 @@ module Yast
     def SecurityLevelHandler(options)
       options = deep_copy(options)
       current = :custom
-      Builtins.maplist(@Levels) do |key, level|
-        current = key if level == Security.Settings
-      end
-      lvl = ""
-      if options.key?("workstation")
-        lvl = "Level1"
-      elsif options.key?("roaming")
-        lvl = "Level2"
-      elsif options.key?("server")
-        lvl = "Level3"
+
+      levels = @Levels.select { |_key, level| level == Security.Settings }
+
+      current = if levels.empty?
+                  :custom
+                else
+                  levels.last # mimic the old implementation
+                end
+
+      lvl = if options.key?("workstation")
+              "Level1"
+            elsif options.key?("roaming")
+              "Level2"
+            elsif options.key?("server")
+              "Level3"
+            else
+              :custom # shouldnt :custom and levels all same type string or symbol?
       end
 
       if current != lvl
-        Security.Settings = Ops.get(@Levels, lvl, {})
+        Security.Settings = @Levels.fetch(lvl, {})
         Security.modified = true
-        return true
       end
-      false
+      Security.modified
     end
 
     # Set value of specific security option
-    # @return [Boolean] false
+    # @return [Boolean] successfully modified?
     def SecuritySetHandler(options)
       options = deep_copy(options)
-      if Builtins.haskey(options, "passwd") &&
-          Ops.get_string(options, "passwd", "") !=
-            Ops.get(Security.Settings, "PASSWD_ENCRYPTION", "")
-        Ops.set(
-          Security.Settings,
-          "PASSWD_ENCRYPTION",
-          Ops.get_string(options, "passwd", "des")
-        )
-        Security.modified = true
-      end
-      if Builtins.haskey(options, "crack") &&
-          Ops.get_string(options, "crack", "") !=
-            Ops.get(Security.Settings, "PASSWD_USE_CRACKLIB", "")
-        Ops.set(
-          Security.Settings,
-          "PASSWD_USE_CRACKLIB",
-          Ops.get_string(options, "crack", "yes")
-        )
-        Security.modified = true
-      end
-      if Builtins.haskey(options, "permissions") &&
-          !Builtins.issubstring(
-            Ops.get(Security.Settings, "PERMISSION_SECURITY", ""),
-            Ops.get_string(options, "permissions", "")
-          )
-        Ops.set(
-          Security.Settings,
-          "PERMISSION_SECURITY",
-          Ops.add(Ops.get_string(options, "permissions", ""), " local")
-        )
+
+      if options.key?("password") &&
+          Security.Settings["PASSWD_ENCRYPTION"] != options["password"]
+        Security.Settings["PASSWD_ENCRYPTION"] = options.fetch("password", "des")
         Security.modified = true
       end
 
-      if Builtins.haskey(options, "remember") &&
-          Ops.get(Security.Settings, "PASSWD_REMEMBER_HISTORY", "0") !=
-            Ops.get_string(options, "remember", "0")
-        to_remember = Builtins.tointeger(
-          Ops.get_string(options, "remember", "0")
-        )
-        if to_remember == nil || Ops.less_than(to_remember, 0) ||
-            Ops.greater_than(to_remember, 400)
-          # error message
+      if options_has_key?("crack") &&
+          Security.Settings["PASSWD_USE_CRACKLIB"] != options["crack"]
+        Security.Settings["PASSWD_USE_CRACKLIB"] = options.fetch("crack", "yes")
+        Security.modified = true
+      end
+
+      if options.key?("permissions") &&
+          Security.Settings["PERMISSION_SECURITY"] != options["permissions"]
+        Security.Settings["PERMISSION_SECURITY"] = options.fetch("permissions", "local")
+        Security.modified = true
+      end
+
+      if options.key?("remember") &&
+          Security.Settings["PASSWD_REMEMBER_HISTORY"] != options["remember"]
+        if options["remember"].to_i.between?(0, 400)
+          Security.Settings["PASSWD_REMEMBER_HISTORY"] = options["remember"].to_i
+          Security.modified = true
+        else
           Report.Error(
             _("The number of passwords to remember must be between 0 an 400.")
           )
-          return false
         end
-        Ops.set(
-          Security.Settings,
-          "PASSWD_REMEMBER_HISTORY",
-          Ops.get_string(options, "remember", "0")
-        )
-        Security.modified = true
+        Security.modified
       end
       Security.modified
     end
--- a/src/clients/security.rb
+++ b/src/clients/security.rb
@@ -167,21 +167,19 @@ module Yast
 
     # Print security summary
     # @return [Boolean] false
-    def SecuritySummaryHandler(options)
-      options = deep_copy(options)
+    def SecuritySummaryHandler(_options)
       sum = Security.Summary
       CommandLine.Print(Ops.get_string(sum, 0, ""))
       false # do not call Write...
     end
 
     # Set security level
-    # @return [Boolean] successfully modified?
+    # @return [Boolean] false
     def SecurityLevelHandler(options)
       options = deep_copy(options)
       current = :custom
 
-      current = @Levels.select{ |_key, level| level == Security.Settings }.keys.last || :custom
-
+      current = @Levels.select { |_key, level| level == Security.Settings }.keys.last || :custom
 
       lvl = if options.key?("workstation")
               "Level1"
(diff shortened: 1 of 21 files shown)


mvidner:
> Hi, I refactored some autoconverted code. I didn't change any functionality. Just refactoring.

Sorry, in the past we have seen that such refactoring often leads to bugs. So it would be good to mention this in package/*.changes.

It appears you have user RuboCop for this. Why not commit also the configuration file?

vpereira:
sure, i used a rubocop configuration used by yast. I will attach it.

src/clients/security.rb
```diff
@@ -180,83 +179,66 @@ def SecuritySummaryHandler(options)
...
-        lvl = "Level2"
-      elsif options.key?("server")
-        lvl = "Level3"
+
+      levels = @Levels.select { |_key, level| level == Security.Settings }
+
+      current = if levels.empty?
+                  :custom
+                else
+                  levels.last # mimic the old implementation
```
jreidinger: this is in fact wrong, as levels contain pair of `[key, level]` and old one use only key.
so how code can look like:
```
current = @Levels.select{ |_key, level| level == Security.Settings }.keys.last || :custom
```

src/clients/security.rb
```diff
@@ -180,83 +179,66 @@ def SecuritySummaryHandler(options)
...
       end
 
       if current != lvl
-        Security.Settings = Ops.get(@Levels, lvl, {})
+        Security.Settings = @Levels.fetch(lvl, {})
         Security.modified = true
-        return true
       end
-      false
+      Security.modified
```
jreidinger: this in fact is not same as previous code as it can return true, when something else set Security.modified to true and here it is not set

src/clients/security.rb
```diff
@@ -180,83 +179,66 @@ def SecuritySummaryHandler(options)
...
           Report.Error(
             _("The number of passwords to remember must be between 0 an 400.")
           )
-          return false
         end
-        Ops.set(
-          Security.Settings,
-          "PASSWD_REMEMBER_HISTORY",
-          Ops.get_string(options, "remember", "0")
-        )
```
jreidinger: this looks like removed without replacemend.

src/clients/security_auto.rb
```diff
@@ -56,30 +56,30 @@ def main
...
       # Check arguments
-      if Ops.greater_than(Builtins.size(WFM.Args), 0) &&
-          Ops.is_string?(WFM.Args(0))
-        @func = Convert.to_string(WFM.Args(0))
-        if Ops.greater_than(Builtins.size(WFM.Args), 1) &&
-            Ops.is_map?(WFM.Args(1))
-          @param = Convert.to_map(WFM.Args(1))
+      if !Yast::WFM.Args.empty?
+        @func = Yast::WFM.Args[0]
+        if Yast::WFM.Args.length > 1 && Yast::WFM.Args[1].is_a?(Hash)
```
jreidinger: in fact second option is enough, because if length is 1 or less, then `Yast::WFM.Args[1]` returns nil and `nil.is_a?(Hash)` returns false

src/clients/security_auto.rb
```diff
@@ -90,37 +90,37 @@ def main
         end
         @ret = Security.Import(
           Map.KeysToUpper(
-            Convert.convert(@param, :from => "map", :to => "map <string, any>")
+            Convert.convert(@param, from: "map", to: "map <string, any>")
```
jreidinger: in fact whole this convert can be dropeed as implementation do not distinct between map and specialized map.

jreidinger: so simple `@ret = Security.Import(Map.KeysToUpper(@param))` can be used

src/include/security/dialogs.rb
```diff
@@ -354,7 +352,7 @@ def DisplayHelpPopup(help_id)
       if help_id == "MANDATORY_SERVICES"
         missing = Security.MissingMandatoryServices
 
-        if missing != nil && missing != []
+        if !missing.nil? && missing != []
```
jreidinger: in fact simple `if missing && !missing.empty?` can be used.

src/include/security/routines.rb
```diff
@@ -112,15 +112,15 @@ def settings2widget(_ID)
         # string|list it
         Builtins.y2debug("li=%1 (%2)", li, i)
         it = Ops.get(li, i)
-        it = "" if it == nil
+        it = "" if it.nil?
```
jreidinger: `it ||= ""`

src/include/security/routines.rb
```diff
@@ -158,11 +158,11 @@ def widget2settings(_ID)
       ret = UI.QueryWidget(Id(_ID), :Value)
       new = ""
       if Ops.is_boolean?(ret)
-        if ret == true
-          new = "yes"
-        else
-          new = "no"
-        end
+        new = if ret == true
```
jreidinger: `new = ret ? "yes" : "no"`

src/modules/Security.rb
```diff
@@ -268,9 +262,9 @@ def PollAbort
     end
 
     # Abort function
-    # @return blah blah lahjk
+    # @return false
     def Abort
-      return Builtins.eval(@AbortFunction) == true if @AbortFunction != nil
+      return Builtins.eval(@AbortFunction) == true if !@AbortFunction.nil?
```
jreidinger: `return Builtins.eval(@AbortFunction) == true if @AbortFunction`

src/modules/Security.rb
```diff
@@ -566,7 +560,11 @@ def write_kernel_settings
       # NOTE: the call to #sort is only needed to satisfy the old testsuite
       @sysctl.sort.each do |key, default_value|
         val = @Settings.fetch(key, default_value)
-        int_val = Integer(val) rescue nil
+        int_val = begin
+                    Integer(val)
+                  rescue
+                    nil
+                  end
```
jreidinger: why is this better then before?

src/modules/Security.rb
```diff
@@ -577,8 +575,12 @@ def write_kernel_settings
       SCR.Write(path(".etc.sysctl_conf"), nil) if written
 
       # enable sysrq?
-      sysrq = Integer(@Settings.fetch("kernel.sysrq", "0")) rescue nil
-      if sysrq != nil
+      sysrq = begin
+                Integer(@Settings.fetch("kernel.sysrq", "0"))
+              rescue
+                nil
+              end
```
jreidinger: same here

src/modules/Security.rb
```diff
@@ -731,35 +733,30 @@ def Import(settings)
     # (For use by autoinstallation.)
     # @return [Hash] Dumped settings (later acceptable by Import ())
     def Export
-      Builtins.eval(@Settings)
+      @Settings if @Settings.is_a?(Hash)
```
jreidinger: well, here is bug, that eval do copy. Also check for hash is not needed, so it should look like
```
deep_copy(@Settings)
```

src/modules/Security.rb
```diff
@@ -731,35 +733,30 @@ def Import(settings)
...
     end
 
     # Create a textual summary and a list of unconfigured cards
     # @return summary of the current configuration
     def Summary
       settings = deep_copy(@Settings)
-      Builtins.foreach(@do_not_test) do |key|
-        settings = Builtins.remove(settings, key)
-      end
+      @do_not_test.each { |k| settings.delete k }
```
jreidinger: what about `settings = settings - @do_not_test` ?

jreidinger: or even better `settings -= @do_not_test`

src/modules/Security.rb
```diff
@@ -731,35 +733,30 @@ def Import(settings)
...
 
       # Summary text
       summary = _("Current Security Level: Custom settings")
       if current != :custom
         # Summary text
-        summary = Builtins.sformat(
-          _("Current Security Level: %1"),
-          Ops.get(@LevelsNames, Convert.to_string(current), "")
-        )
+        summary = _("Current Security Level: #{@LevelsNames[current]}")
```
jreidinger: no, this won't work, as `_` method get extrapolated string, which is wrong. It have to look in ruby like
`summary = _("Current Security Level: %s") % @LevelsNames[current]`

src/modules/Security.rb
vpereira: proposed by the rubycop :)

src/include/security/routines.rb
```diff
@@ -89,7 +89,7 @@ def settings2widget(_ID)
       # "Widget" == "IntField"
       if widget == "IntField"
         intval = Builtins.tointeger(value)
-        intval = 0 if intval == nil
+        intval = 0 if intval.nil?
```
vpereira: proposed by the rubycop :)

jreidinger: I know, as my proposal won't work if intval will be boolean, but it is not, so it is safer and nicer.

src/clients/security_auto.rb
```diff
@@ -130,10 +120,6 @@ def main
       Builtins.y2debug("ret=%1", @ret)
       Builtins.y2milestone("Security auto finished")
       Builtins.y2milestone("----------------------------------------")
-
-      # is it needed?
-      deep_copy(@ret)
-      # EOF
```
jreidinger: this is in fact wrong as you kill return value. I just said that deep_copy is not needed, so still you need to have here `@ret` otherwise it return nil, causing bugs.

src/include/security/routines.rb
```diff
@@ -112,7 +112,7 @@ def settings2widget(_ID)
         # string|list it
         Builtins.y2debug("li=%1 (%2)", li, i)
         it = Ops.get(li, i)
-        it = "" if it.nil?
+        it = ||= ""
```
jreidinger: this is wrong, it should be `it ||= `

src/modules/Security.rb
vpereira: hm.. settings is a hash, no? Plain hashes don't have the "-" method implemented, I'm not sure if I can do it.

jreidinger: ok, if it is Hash, then delete is right approach

src/clients/security.rb
vpereira: I changed it already to use strings. I think there is no problem, but consistence is good.

src/clients/security.rb
vpereira: and why we return always false? there is a reason for that?

jreidinger: I expect that logic is that we return false to indicate that this action do not change anything.

vpereira:
I fixed almost all issues. Rubocop has just three complains related with useless assignment. Could you please evaluate them? 

```
src/clients/security.rb:180:7: W: Lint/UselessAssignment: Useless assignment to variable - current. 
      current = :custom
      ^^^^^^^
src/include/security/complex.rb:138:9: W: Lint/UselessAssignment: Useless assignment to variable - cur. 
        cur = UI.QueryWidget(Id(:rb), :CurrentButton)
        ^^^
src/include/security/dialogs.rb:289:9: W: Lint/UselessAssignment: Useless assignment to variable - current_value. 
        current_value = Ops.get(
        ^^^^^^^^^^^^^
```

jreidinger:
@vpereira OK, I evaluated them and result is:

- src/clients/security.rb:180 - valid as you set custom few lines lower as default, so please remove it

- src/include/security/complex.rb:138 - also valid as when cur is used, it is few lines below, so please remove it

- src/include/security/dialogs.rb:289 - also valid, looks like old code, now it uses label_mapping for label instead of name, so also can be removed

src/clients/security.rb
```diff
@@ -168,34 +167,32 @@ def main
...
+              "Level3"
+            else
+              :custom # shouldnt :custom and levels all same type string or symbol?
+            end
 
       if current != lvl
-        Security.Settings = Ops.get(@Levels, lvl, {})
+        Security.Settings = @Levels.fetch(lvl, {})
         Security.modified = true
-        return true
```
jreidinger: is it valid? what is reason for this change?

src/lib/security/display_manager.rb
```diff
@@ -19,14 +19,15 @@
 # current contact information at www.suse.com.
 # ------------------------------------------------------------------------------
 #
-
+# rubocop:disable Style/MutableConstant
```
jreidinger: why it is needed? why not simply freeze array?

jreidinger: ah, below we in fact modify it. It does not look correct for me.

src/modules/Security.rb
```diff
@@ -64,19 +63,18 @@ def import_modules
...
-      if srv_file
-        srv_lists = YAML.load_file(srv_file) rescue {}
-      else
-        srv_lists = {}
-      end
+      srv_lists = srv_file ? (begin
+                                YAML.load_file(srv_file)
+                              rescue
+                                {}
+                              end) : {}
```
jreidinger: to be honest this looks horrible. why not use
```
srv_lists = {}
if srv_file
  begin
     srv_lists = YAML.load_file(srv_file)
  rescue => e
    log.warn "Failed to load yaml file: #{e.message}"
  end
end
```

jreidinger:
ok, I added few mote notes, that looks wrong for me. Otherwise it looks good.

src/modules/Security.rb
vpereira: proposed by rubycop ;)

lslezak:
Ping, @vpereira could you look at the @jreidinger 's comments?

vpereira:
hi, I will check it ASAP, sorry for the delay

src/modules/Security.rb
vpereira: thats matter of taste. I can use your approach. However I prefer the "functional" style returning and assigning the value of if.

vpereira: i'm doing as you suggested @jreidinger

teclator:
@vpereira @jreidinger it seems this PR is just waiting for the missing answer about a change or returned values.

@vpereira do you plan to work on it ? if not, as it is only a cleanup and no bug related I would close it until we are really able to finish it

vpereira:
hi  @teclator , the communication in this PR didn't work so well.. some long period of silence between the commits makes hard for someone sporadically contributing to the codestream to come back and work on that. There are still a question from my side, without answer. Beside it, maybe some commits should be cherry-picked, since it was started in 2016 and probably we updated already our rubocop rules.. Please advice what should be done here.

teclator:
@vpereira up to you. I mean, you can rebase with master fixing the conflicts and check rubocop. If you do not have time or do not plan to work on it then taking in account that we are in RC phase and that it is just a cleanup and do not fix any bug at all I would close it.

PR #47, opened 2018-11-02
Writing security settings in first AY installation stage.

schubi2:


src/clients/security_finish.rb
```diff
@@ -0,0 +1,9 @@
+# encoding: utf-8
+
+# File:	clients/security_finish.ycp
+# Package: Security configuration
+# Summary: Configuration of /etc/login.defs /etc/login.defs
```
lslezak: I forgot what all this module does, is it still valid? Is this up to date?

PR #54, opened 2019-09-05
Sle 15 ga bsc 1147173

schubi2:


src/modules/Security.rb
```diff
@@ -395,21 +395,18 @@ def read_pam_settings
...
-              else
-                "secure"
-              end
-
-      @Settings["PERMISSION_SECURITY"] = perm
-
-      log.debug "PERMISSION SECURITY (after #{__callee__}): " \
+      # Removing "local" from the string
+      permissions = @Settings["PERMISSION_SECURITY"].to_s.split(" ")
+      @Settings["PERMISSION_SECURITY"] = permissions.delete_if {|p|
```
dgdavid: If it does not fit in a single line, I'd write it as a `do`/`end` block.

dgdavid: Even more, maybe this could be done simply with a `gsub`, isn't it?

```ruby
permissions = @Settings["PERMISSION_SECURITY"].to_s.gsub("local", "").gsub(/\s+/, " ")
@Settings["PERMISSION_SECURITY"] = permissions.empty? ? "secure" : permissions
```

Just an idea.

test/security_test.rb
```diff
@@ -477,16 +477,23 @@ def enabled?
...
-          Security.Settings["PERMISSION_SECURITY"] = "it_is_easy_to_test"
+          Security.Settings["PERMISSION_SECURITY"] = "easy local"
 
           expect(Security.read_permissions).to eql("easy")
           expect(Security.Settings["PERMISSION_SECURITY"]).to eql("easy")
         end
 
-        it "sets security permission to 'paranoid' if contains paranoid" do
-          Security.Settings["PERMISSION_SECURITY"] = "paranoid_permission"
+        it "sets user defined security permission " do
```
dgdavid: NP: extra whit space at the end of the sentence

src/modules/Security.rb
schubi2: Maybe I am old fashioned, but I do not really like the pattern stuff together with gsub. For me it is harder to understand. But maybe I am wrong.....
And somehow it does not really work regarding white spaces.
     Failure/Error: expect(Security.read_permissions).to eql("user_defined")
     
       expected: "user_defined"
            got: " user_defined "
     
       (compared using eql?)
     # ./security_test.rb:491:in `block (4 levels) in <module:Yast>'

  3) #<Yast::SecurityClass:0x0000000002094b90> #read_permissions depending on current permission removes local permission
     Failure/Error: expect(Security.read_permissions).to eql("paranoid")
     
       expected: "paranoid"
            got: "paranoid "

dgdavid: Nah, let keep it as you implemented. There is a missing `.strip` and it could be improved extracting the "sanitize" logic to another method. But, I don't want to block this.

Just go ahead. Sorry for being a nitpicker :)

PR #57, opened 2019-09-27
Sle 15 sp1 bsc 1150821

schubi2:


src/modules/Security.rb
```diff
@@ -314,13 +314,7 @@ def inittab_shutdown_configured?
...
     # See bug 742783 for description
     def ReadConsoleShutdown
-      ret = ::Security::CtrlAltDelConfig.current || ::Security::CtrlAltDelConfig.default
-
-      return ret if ::Security::CtrlAltDelConfig.systemd?
-
-      @Settings["CONSOLE_SHUTDOWN"] = ret if ::Security::CtrlAltDelConfig.inittab?
-
-      nil
+      ::Security::CtrlAltDelConfig.current || ::Security::CtrlAltDelConfig.default
```
teclator: Why you discarded this solution?.

It could be something like:

```ruby
def ReadConsoleShutdown
  @Settings["CONSOLE_SHUTDOWN"] = ::Security::CtrlAltDelConfig.current || ::Security::CtrlAltDelConfig.default

  nil
end
```

And in Read just modify:
```ruby
@Settings["CONSOLE_SHUTDOWN"] = ReadConsoleShudtown()
```
by
```ruby
ReadConsoleShudtown()
```

... just to know the reasons behind the change.

src/modules/Security.rb
```diff
@@ -730,6 +730,10 @@ def Import(settings)
     # (For use by autoinstallation.)
     # @return [Hash] Dumped settings (later acceptable by Import ())
     def Export
```
teclator: For me, it is somehow strange that Export modify the internal state of Security module. It should only export it.

test/security_test.rb
```diff
@@ -700,5 +700,21 @@ def enabled?
       end
 
     end
+
+    describe "#Export" do
+      before do
+        Security.Settings["FAIL_DELAY"] = "5"
```
teclator: These variables looks like unnecessary or at least do not see tests for them.

src/lib/security/ctrl_alt_del_config.rb
```diff
@@ -27,14 +27,14 @@ module CtrlAltDelConfig
...
-    Yast.import "Package"
+    Yast.import "PackageSystem"
     Yast.import "FileUtils"
 
     SYSTEMD_FILE = "/etc/systemd/system/ctrl-alt-del.target"
 
     class << self
       def systemd?
-        Yast::Package.Installed("systemd")
+        Yast::PackageSystem.Installed("systemd")
```
teclator: I guess this change does not affect any other scenario, isn't it?

schubi2: I think it is quite save because it calls rpm directly.

PR #59, opened 2019-10-03
Write sysctl settings to /etc/sysctl.d

imobachgs:
Adapts sysctl handling to use the new [Yast2::CFA::Sysctl](https://github.com/yast/yast-yast2/pull/966) class so the configuration is written to a YaST specific file under `/etc/sysctl.d`.

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -27,7 +27,7 @@
 # $Id$
 require "yast"
 require "yast2/systemd/service"
-require "yast2/cfa/sysctl"
+require "cfa/sysctl"
 require "yaml"
 require "security/ctrl_alt_del_config"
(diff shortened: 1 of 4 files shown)


src/modules/Security.rb
```diff
@@ -27,6 +27,7 @@
 # $Id$
 require "yast"
 require "yast2/systemd/service"
+require "yast2/cfa/sysctl"
```
jreidinger: hmm, I just see it, why not it live in `cfa/sysctl`? My idea behind living it in global cfa namespace is to make it easier in future to separate it to own gem if we decide.

imobachgs: Yes, no problem. I will move it.

PR #61, opened 2019-11-22
Change default encrypt method

imobachgs:
## Problem

After moving the `/etc/login.defs` to `/usr/etc/login.defs`, YaST2 security does not find the current configuration. The problem is that the fallback encryption method is `des`, which causes users/groups passwords to use that method instead of `sha512`.

See https://bugzilla.suse.com/show_bug.cgi?id=1157541

## Solution

Use `sha512` as default. In the screenshot below you can see the `/etc/shadow` content and the security settings screen after applying the patch.

yast2-security

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -35,6 +35,7 @@ require "security/display_manager"
 module Yast
   class SecurityClass < Module
     DEFAULT_ENCRYPT_METHOD = "sha512".freeze
+    private_constant :DEFAULT_ENCRYPT_METHOD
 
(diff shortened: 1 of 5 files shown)


src/modules/Security.rb
```diff
@@ -34,6 +34,7 @@
 
 module Yast
   class SecurityClass < Module
+    DEFAULT_ENCRYPT_METHOD = "sha512".freeze
```
dgdavid: NP: As you added a public method to expose it, do you think it should be a `private_constant`?

PR #62, opened 2019-11-22
Read login.defs settings from /usr/etc

imobachgs:
See https://github.com/yast/yast-yast2/pull/988.

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -881,7 +881,7 @@ module Yast
     end
 
     # @param key    [String] Key to set the value for
-    # @return value [String] Value to assign to the given key
+    # @param value [String] Value to assign to the given key
     def write_sysctl_value(key, value)
       sysctl_file.public_send(SYSCTL_KEY_TO_METH[key].to_s + "=", value)
     end
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -28,6 +28,7 @@
 require "yast"
 require "yast2/systemd/service"
 require "cfa/sysctl"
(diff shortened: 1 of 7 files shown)


test/security_test.rb
```diff
@@ -155,13 +159,23 @@ def enabled?
...
+        expect(was_written?(".sysconfig.clock")).to eq(true)
+      end
+    end
+
+    describe "#write_shadow_config" do
+      before do
+        Security.Settings["FAIL_DELAY"] = "10"
+      end
+
+      it "reads login.defs configuration" do
```
dgdavid: s/reads/writes

src/modules/Security.rb
```diff
@@ -517,14 +534,18 @@ def write_to_locations
...
-      # pam stuff
-      encr = @Settings.fetch("PASSWD_ENCRYPTION", default_encrypt_method)
-      if encr != @Settings_bak["PASSWD_ENCRYPTION"]
-        SCR.Write(path(".etc.login_defs.ENCRYPT_METHOD"), encr)
+    # Write login.defs configuration
+    def write_shadow_config
+      SHADOW_ATTRS.each do |attr|
+        shadow_config.public_send("#{attr.to_s.downcase}=", @Settings[attr])
       end
+      encr = @Settings.fetch("PASSWD_ENCRYPTION", default_encrypt_method)
```
dgdavid: For parity, I'd have a `write_encryption_method` method. But I understand that it implies calling `#save` twice.

imobachgs: Yes, there is some asymmetry there.

PR #66, opened 2020-02-03
Using SysctlConfig class

schubi2:
Handling Sysctl entries in following files/directories:
/run/sysctl.d,
/etc/sysctl.d
/usr/local/lib/sysctl.d
/usr/lib/sysctl.d
/lib/sysctl.d
/etc/sysctl.conf

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -730,16 +730,16 @@ module Yast
     def Import(settings)
       settings = deep_copy(settings)
       if settings.key?("KERNEL.SYSRQ")
-        settings["kernel.sysrq"] = settings.delete["KERNEL.SYSRQ"]
+        settings["kernel.sysrq"] = settings.delete("KERNEL.SYSRQ")
       end
       if settings.key?("NET.IPV4.TCP_SYNCOOKIES")
-        settings["net.ipv4.tcp_syncookies"] = settings.delete["NET.IPV4.TCP_SYNCOOKIES"]
+        settings["net.ipv4.tcp_syncookies"] = settings.delete("NET.IPV4.TCP_SYNCOOKIES")
       end
       if settings.key?("NET.IPV4.IP_FORWARD")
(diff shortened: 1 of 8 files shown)


src/modules/Security.rb
```diff
@@ -726,6 +729,19 @@ def Write
     # @return [Boolean] True on success
     def Import(settings)
       settings = deep_copy(settings)
+      if settings.key?("KERNEL.SYSRQ")
```
imobachgs: Why are you fixing the 'case' of these attributes only? What about the rest? On the other hand, the hash will have some "duplicated" values:

```ruby
{
  "KERNEL.SYSRQ" => "1",
  "kernel.sysrq" => "1"
}
```

It will work, but it might be confusing. You can avoid this by using: `#delete`:

```ruby
settings["kernel.sysrq"] = settings.delete("KERNEL.SYSRQ")
```

src/modules/Security.rb
```diff
@@ -583,14 +583,17 @@ def write_kernel_settings
       @sysctl.sort.each do |key, default_value|
         val = @Settings.fetch(key, default_value)
         int_val = Integer(val) rescue nil
-        if int_val.nil?
-          log.error "value #{val} for #{key} is not integer, not writing"
+        if int_val.nil? && ![TrueClass, FalseClass].include?(val.class)
```
imobachgs: np: Why not simply:

```ruby
int_val.nil? && ![true, false].include?(val)
```

src/modules/Security.rb
```diff
@@ -750,6 +771,17 @@ def Import(settings)
     # (For use by autoinstallation.)
     # @return [Hash] Dumped settings (later acceptable by Import ())
     def Export
+      # conversion to 0/1 string
+      if [TrueClass, FalseClass].include?(@Settings["net.ipv4.ip_forward"].class)
```
imobachgs: np: Same as above. Why not simply `[true, false].include?(...)`.

PR #70, opened 2020-07-29
Use pam_pwquality instead of pam_cracklib depending on availability

aschnell:
The PR makes YaST select between pam_pwquality and pam_cracklib depending on availability.

For https://trello.com/c/ahoBngck/1948-5-ostumbleweed-p5-1171318-build-20200506-yast-security-module-integrate-with-pampwdquality and https://bugzilla.suse.com/show_bug.cgi?id=1171318. Also fixes https://bugzilla.suse.com/show_bug.cgi?id=1174619.

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -126,7 +126,7 @@ module Yast
         "GID_MIN"                                   => "1000",
         "HIBERNATE_SYSTEM"                          => "active_console",
         "PASSWD_ENCRYPTION"                         => "sha512",
-        "PASSWD_USE_CRACKLIB"                       => "yes",
+        "PASSWD_USE_PWQUALITY"                      => "yes",
         "PASS_MAX_DAYS"                             => "99999",
         "PASS_MIN_DAYS"                             => "0",
         "PASS_MIN_LEN"                              => "5",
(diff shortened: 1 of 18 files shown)


src/modules/Security.rb
```diff
@@ -545,24 +555,24 @@ def write_shadow_config
...
+      if @Settings["PASSWD_USE_PWQUALITY"] == "yes"
+        Pam.Add(pwquality_module)
         pth = @Settings["CRACKLIB_DICT_PATH"]
         if pth && pth != "/usr/lib/cracklib_dict"
-          Pam.Add("--cracklib-dictpath=#{pth}")
+          Pam.Add(pwquality_module + "-dictpath=#{pth}")
         end
       else
-        Pam.Remove("cracklib")
+        Pam.Remove(pwquality_module)
```
jreidinger, line 566: here I am not sure if it is correct. What happens when user use old cracklib and now new pwquality is available? I think it is safer to remove both here, not?

aschnell: Well, switching from pam_cracklib to pam_pwquality will be the task of pam. YaST is not involved here.

Anyway, already now adding cracklib adds pwquality (if pam-deprecated is installed). I was not able to add cracklib anymore so removing it seems useless to me.

PR #75, opened 2020-10-27
Translate <enable_sysrq> to <kernel.sysrq> with the correct value

schubi2:
AY-Import: Translate <enable_sysrq> setting to <kernel.sysrq> with the correct value format (bsc#1177720).

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -41,9 +41,13 @@ module Yast
     include Yast::Logger
     include ::Security::CtrlAltDelConfig
 
-    SYSCTL_VALUES = {
+    SYSCTL_VALUES_TO_BOOLEAN = {
       "yes" => true,
       "no"  => false
+    }
+    SYSCTL_VALUES_TO_INTSTRING = {
+      "yes" => "1",
+      "no"  => "0"
     }
 
     SHADOW_ATTRS = [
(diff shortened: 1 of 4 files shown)


src/modules/Security.rb
```diff
@@ -762,9 +766,15 @@ def Import(settings)
           tmpSettings[k] = settings[k]
         else
           if @sysctl.key?(k) && settings.key?(@sysctl2sysconfig[k])
+            # using the old sysconfig AY format
             val = settings[@sysctl2sysconfig[k]].to_s
-            tmpSettings[k] = SYSCTL_VALUES.key?(val) ? SYSCTL_VALUES[val] : val
```
teclator, line 766: It looks ok but it seems tere are some duplication of code, in previous lines some keys are treating explicitly and then we do the transformation or translation of keys. Maybe we could move the translation before the specific code simplifying.

https://github.com/yast/yast-security/blob/da1e605aa17229b59a0b5aaf6045d1cb2ddfcbaf/src/modules/Security.rb#L740-L758

schubi2: Hm, for me it is completely unrelated. I must admit that on the first view I have had the same thoughts :-)
Here the the new kernel settings will be lower cased only in order to get an unified key.
Around my fix a translation will be done from old format to newer one.

teclator (approved):
LGTM

PR #79, opened 2020-12-15
Removed old code for sysvinit configuration (bsc#1175494).

schubi2:
Related to https://bugzilla.suse.com/show_bug.cgi?id=1175494, there are  some fallback to sysvinit that can be removed.


dgdavid (approved):
Looks good, thanks :)

PR #82, opened 2021-01-29
WIP: Security::Selinux

dgdavid:
Just a draft PR to discuss the desired Security::Selinux API.

### DIFF
--- a/src/lib/security/selinux.rb
+++ b/src/lib/security/selinux.rb
@@ -26,7 +26,7 @@ module Security
     include Yast::Logger
     Yast.import "Bootloader"
 
-    GETENFORCE_PATH = "/usr/sbin/getenforce"
+    GETENFORCE_PATH = "/usr/sbin/getenforce".freeze
     private_constant :GETENFORCE_PATH
 
     class << self
--- a/src/lib/security/selinux.rb
+++ b/src/lib/security/selinux.rb
@@ -60,7 +60,7 @@ module Security
 
     # Returns the policy set in booting params
     def policy
-      @policy ||= match_policy(policy_from_kernel_params) || :disabled
+      @policy ||= match_policy(policy_options_from_kernel) || :disabled
     end
 
     # Use the given policy for the next boot (in running system is needed to #save)
@@ -108,7 +108,7 @@ module Security
       self.class.policies.key(policy_options)
     end
 
-    def policy_from_kernel_params
+    def policy_options_from_kernel
       Hash[*self.class.policy_keys.flat_map { |key| [key, read_param(key)] }]
     end
 
--- a/src/lib/security/selinux.rb
+++ b/src/lib/security/selinux.rb
@@ -29,23 +29,30 @@ module Security
     GETENFORCE_PATH = "/usr/sbin/getenforce"
     private_constant :GETENFORCE_PATH
 
-    POLICY_KEYS = ["security", "selinux", "enforcing"]
-    private_constant :POLICY_KEYS
+    class << self
+      DEFAULT_POLICY_OPTIONS = {
+        "security"  => :missing,
+        "selinux"   => :missing,
+        "enforcing" => :missing
+      }.freeze
+      private_constant :DEFAULT_POLICY_OPTIONS
 
-    ENABLE_POLICY = { "security" => "selinux", "selinux" => "1" }
-    private_constant :ENABLE_POLICY
+      def define_policy(name, options = {})
+        policies[name] = DEFAULT_POLICY_OPTIONS.merge(options)
+      end
 
-    DISABLE_POLICY = Hash[*POLICY_KEYS.flat_map { |key| [key, :missing] }]
-    private_constant :DISABLE_POLICY
+      def policies
+        @policies ||= {}
+      end
 
-    POLICIES = {
-      :disabled   => DISABLE_POLICY,
-      :permissive => ENABLE_POLICY.merge("enforcing" => :missing),
-      :enforcing  => ENABLE_POLICY.merge("enforcing" => "1")
-    }
-    private_constant :POLICIES
+      def policy_keys
+        DEFAULT_POLICY_OPTIONS.keys
+      end
+    end
 
-    attr_reader :changed
+    define_policy(:disabled)
+    define_policy(:permissive, "security" => "selinux", "selinux" => "1", "enforcing" => :missing)
+    define_policy(:enforcing, "security" => "selinux", "selinux" => "1", "enforcing" => "1")
 
     def initialize
       propose_policy if Yast::Mode.installation
@@ -53,15 +60,17 @@ module Security
 
     # Returns the policy set in booting params
     def policy
-      POLICIES.key(policy_from_kernel_params) || :disabled
+      @policy ||= match_policy(policy_from_kernel_params) || :disabled
     end
 
     # Use the given policy for the next boot (in running system is needed to #save)
     def policy=(key)
-      if POLICIES.keys.include?(key)
-        log.debug("Changing SELinux to #{key} mode: #{POLICIES[key]}")
+      found_policy = find_policy(key)
 
-        update_policy(*POLICIES[key])
+      if found_policy
+        log.debug("Changing SELinux to #{key} mode: #{found_policy}")
+
+        @policy = key
       else
         log.debug("Unknown `#{key}` SELinux policy")
       end
@@ -78,26 +87,38 @@ module Security
     def propose_policy
       key = :enforcing # read it from control-file
 
-      update_policy(POLICIES[key])
-      key
+      log.debug "Proposing the `#{key}` SELinux policy: #{@policy}"
+      @policy = key
+      save
     end
 
     def save
-      Yast::Bootloader.Write
+      update_policy
+
+      Yast::Bootloader.Write unless Yast::Mode.installation
     end
 
     private
 
+    def find_policy(key)
+      self.class.policies[key]
+    end
+
+    def match_policy(policy_options)
+      self.class.policies.key(policy_options)
+    end
+
     def policy_from_kernel_params
-      Hash[*POLICY_KEYS.flat_map { |key| [key, read_param(key)] }]
+      Hash[*self.class.policy_keys.flat_map { |key| [key, read_param(key)] }]
     end
 
     def read_param(key)
       Yast::Bootloader.kernel_param(:common, key.to_s)
     end
 
-    def update_policy(*params)
-      Yast::Bootloader.modify_kernel_params(*params)
+    def update_policy
+      policy = find_policy(@policy)
+      Yast::Bootloader.modify_kernel_params(**policy)
     end
   end
 end
--- a/src/lib/security/selinux.rb
+++ b/src/lib/security/selinux.rb
@@ -54,51 +54,64 @@ module Security
     define_policy(:permissive, "security" => "selinux", "selinux" => "1", "enforcing" => :missing)
     define_policy(:enforcing, "security" => "selinux", "selinux" => "1", "enforcing" => "1")
 
+    attr_reader :policy
+
     def initialize
       propose_policy if Yast::Mode.installation
-    end
-
-    # Returns the policy set in booting params
-    def policy
-      @policy ||= match_policy(policy_options_from_kernel) || :disabled
-    end
-
-    # Use the given policy for the next boot (in running system is needed to #save)
-    def policy=(key)
-      found_policy = find_policy(key)
-
-      if found_policy
-        log.debug("Changing SELinux to #{key} mode: #{found_policy}")
-
-        @policy = key
-      else
-        log.debug("Unknown `#{key}` SELinux policy")
-      end
+      @policy = configured_policy
     end
 
     # Returns the policy applied in the running system
     def running_policy
       Yast::Execute.locally!(GETENFORCE_PATH, stdout: :capture).chomp.downcase.to_sym
     rescue Cheetah::ExecutionFailed
-      nil
+      :disabled
+    end
+
+    # If current policy value is different to the configured one
+    def changed?
+      @policy != configured_policy
+    end
+
+    # Use the given policy for the next boot (in running system is needed to #save)
+    def update_policy(policy_key)
+      found_policy = find_policy(policy_key)
+
+      if found_policy
+        log.debug("Changing SELinux configuration to `#{policy_key}` policy: #{found_policy}")
+
+        @policy = policy_key
+      else
+        log.debug("Unknown `#{policy_key}` SELinux policy")
+      end
+    end
+    alias_method :policy=, :update_policy
+
+    def save
+      policy_options = find_policy(policy)
+
+      return unless policy_options && changed?
+
+      log.debug("Writting SELinux kernel params: #{policy_options}")
+
+      Yast::Bootloader.modify_kernel_params(**policy_options)
+      Yast::Bootloader.Write unless Yast::Mode.installation
+    end
+
+    private
+
+    def configured_policy
+      match_policy(policy_options_from_kernel) || :disabled
     end
 
     # Propose a policy according to the value set in the control-file
     def propose_policy
       key = :enforcing # read it from control-file
 
-      log.debug "Proposing the `#{key}` SELinux policy: #{@policy}"
-      @policy = key
+      log.debug "Proposing the `#{key}` SELinux policy"
+      update_policy(policy_key)
       save
     end
-
-    def save
-      update_policy
-
-      Yast::Bootloader.Write unless Yast::Mode.installation
-    end
-
-    private
 
     def find_policy(key)
       self.class.policies[key]
@@ -117,8 +130,20 @@ module Security
     end
 
     def update_policy
-      policy = find_policy(@policy)
-      Yast::Bootloader.modify_kernel_params(**policy)
+      found_policy = find_policy(@policy)
+
+      if found_policy
+        log.debug("Changing SELinux to #{key} mode: #{found_policy}")
+
+        @policy = key
+      else
+        log.debug("Unknown `#{key}` SELinux policy")
+      end
+
+      if policy
+        Yast::Bootloader.modify_kernel_params(**policy)
+      else
+      end
     end
   end
 end
--- a/src/lib/security/selinux.rb
+++ b/src/lib/security/selinux.rb
@@ -1,0 +1,103 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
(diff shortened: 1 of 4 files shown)


src/lib/security/selinux.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    def save
+      Yast::Bootloader.Write
+    end
+
+    private
+
+    def param(key)
+      Yast::Bootloader.kernel_param(:common, key.to_s)
```
jreidinger, line 82: be awre that kernel parameter can be also array. I am not sure how it will behave in selinux case if something is specified multiple times.
see https://github.com/config-files-api/config_files_api_grub2/blob/master/lib/cfa/grub2/default.rb#L217

I have to update also Bootloader module documentation. It was added when kernel has specific multiple output consoles.

src/lib/security/selinux.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    def permissive?
+      return false unless enabled?
+
+      !enforcing?
+    end
+
+    def disabled?
+      !enabled?
+    end
```
jreidinger, line 51: what I am missing there is method like `policy` or `value` which can return any value from  list `[:permissive, :enforcing, :disabled]` So it will be easier to get current value for widget. Otherwise I have to ask for each one.

Also what would be useful is to have human name mapping to allow display of value like https://github.com/yast/yast-installation/pull/906/files#diff-21c7a291eed8a2f9df01ced4faea7e8be674002dc095aa746f12f75c018014fcR144

src/lib/security/selinux.rb
```diff
@@ -0,0 +1,89 @@
...
+module Security
+  # Class for handling SELinux kernel params
+  class Selinux
+    Yast.import "Bootloader"
+
+    MODULE_NAME = "selinux".freeze
+    private_constant :MODULE_NAME
+
+    ENABLE_VALUE = "1".freeze
+    private_constant :ENABLE_VALUE
```
jreidinger, line 31: just note: for installation we need to somehow detect that class is run first time and in such case do proposal. Probably in constructor in Mode.initial?

src/lib/security/selinux.rb
jreidinger: and when we have that method value or policy, it would be nice to have also writter, so value from combobox can be directly just store in widget and do not need to check value and call respective method.

src/lib/security/selinux.rb
dgdavid: If I understood it correctly yesterday, [`Yast::Bootloader.ReadOrProposeIfNeeded`](https://github.com/yast/yast-bootloader/blob/f953fef1cb876d8a7229fee0991a443fb3bd5dcf/src/modules/Bootloader.rb#L441) will do the proposal. Doesn't it?

That method is internally executed in both methods used here, [`Yast::Bootloader.kernel_param`](https://github.com/yast/yast-bootloader/blob/f953fef1cb876d8a7229fee0991a443fb3bd5dcf/src/modules/Bootloader.rb#L332) and [`Yast::Bootloader.modify_kenel_paraams`](https://github.com/yast/yast-bootloader/blob/f953fef1cb876d8a7229fee0991a443fb3bd5dcf/src/modules/Bootloader.rb#L376). So, the proposal will be done anyway. Am I wrong?

dgdavid:
@jreidinger 

Making more manual testing, I realized that we have (at least) two scenarios:

* The _desired_ SELinux settings at boot time
* The _chosen_ SELinux settings at boot time (i.e., the user edited the bootloader entry before booting)

Thus, relying in `Yast::Bootloader.kernel_param` is not enough to know the SELinux _real_ status in a running system. We should use `sestatus` or `getenforce` too. Thoughts?

| Editing the Grub entry | Checking SELinux status |
| - | - |
| Screenshot_opensusetumbleweed_2021-01-29_11:36:32 | Screenshot_opensusetumbleweed_2021-01-29_11:41:13 |

src/lib/security/selinux.rb
jreidinger: it is proposal for boot parameters. But I am talking about selinux proposal ( similar like kdump proposal ), which according to trello card should depends on control file?

dgdavid: Ah! right!

dgdavid:
@jreidinger 

New approach, now having `#policy`, `#policy=`, and `#running_policy` methods. Latest one just in case we needed to know the _active_ policy in a running system.

A small manual test

```ruby
require "./src/lib/security/selinux.rb"
@selinux = Security::Selinux.new

puts "SELinux configured policy"
puts "-------------------------"
puts @selinux.policy
puts

puts "How SELinux kernel params looks like"
puts "------------------------------------"
puts @selinux.send(:policy_from_kernel_params)
puts

puts "getenfoce output"
puts "----------------"
puts Yast::Execute.locally!("/usr/sbin/getenforce", stdout: :capture)
puts

puts "SELinux running policy"
puts "----------------------"
puts @selinux.running_policy
puts

```

gives me good results 

```
# irb -r "./test.rb"

SELinux configured policy
-------------------------
enforcing

How SELinux kernel params looks like
------------------------------------
{"security"=>"selinux", "selinux"=>"1", "enforcing"=>"1"}

getenfoce output
----------------
Disabled

SELinux running policy
----------------------
disabled
```

How do you see it now?

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,149 @@
...
+require "yast2/execute"
+
+module Security
+  # Class for handling SELinux kernel params
+  class SelinuxConfig
+    include Yast::Logger
+    Yast.import "Bootloader"
+
+    GETENFORCE_PATH = "/usr/sbin/getenforce".freeze
+    private_constant :GETENFORCE_PATH
```
jreidinger, line 30: it probably just matter of taste but I usually place private constants to private section of class or at the bottom if private section is missing as usually the top of class contain some intro or high level documentation and is often read. So private stuff I usually place at the bottom or in private section.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,149 @@
...
+      end
+
+      def policy_keys
+        DEFAULT_POLICY_OPTIONS.keys
+      end
+    end
+
+    define_policy(:disabled)
+    define_policy(:permissive, "security" => "selinux", "selinux" => "1", "enforcing" => :missing)
+    define_policy(:enforcing, "security" => "selinux", "selinux" => "1", "enforcing" => "1")
```
jreidinger, line 55: to be honest I am not sure if this is the best way to define policies as it is spread over multiple places. Just consider if we want to add new policy where we need to add it. I prefer to have simple map of policies as private constant that explicitelly map policy with kernel options. E.g. here it confuse me why disabled does not have anything and why permissive has enforcing missing when it is default option. I am also not sure about that security => :missing as it can also mean that by disabling selinux you disable also if user define apparmor as security module.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    private
+
+    def configured_policy
+      match_policy(policy_options_from_kernel) || :disabled
+    end
+
+    # Propose a policy according to the value set in the control-file
+    def propose_policy
+      key = :enforcing # read it from control-file
```
jreidinger, line 109: I would use TODO: in comment to not forget it

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      log.debug "Proposing the `#{key}` SELinux policy"
```
jreidinger, line 111: I would use the info instead of debug here. It is important info for us.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    def policy_options_from_kernel
+      Hash[*self.class.policy_keys.flat_map { |key| [key, read_param(key)] }]
+    end
+
+    def read_param(key)
+      Yast::Bootloader.kernel_param(:common, key.to_s)
+    end
+
+    def update_policy
```
jreidinger, line 132: twice method update_policy is bad idea.

dgdavid: Sure, it is. This was just a leftover at the time to _polish_ the draft. Good catch, anyway.

src/lib/security/selinux_config.rb
dgdavid: > I am also not sure about that security => :missing as it can also mean that by disabling selinux you disable also if user define apparmor as security module.

That's a good question. I'm not sure either. But I'm not sure even if at this point we can know if disabling SELinux could mean enabling AppArmor.

dgdavid: > I prefer to have simple map of policies as private constant that explicitelly map policy with kernel options.

Actually I started with a [similar approach](https://github.com/yast/yast-security/pull/82/commits/5e9e3d120854b0a841d8fa96079d966974f0aa63) but I found this other way more _semantic_ and elegant. But let's _rollback_ to a _POLICIES_ map as you suggest.

dgdavid: > But let's _rollback_ to a _POLICIES_ map as you suggest.

Ready. See https://github.com/yast/yast-security/pull/82/commits/9d11f6778f07e64d186df9d58beb787f79bcc743

dgdavid: > But I'm not sure even if at this point we can know if disabling SELinux could mean enabling AppArmor.

I tried just removing the `selinux=1` to see if it is enough to disable SELinux, without touching `security=selinux` but it does not work. The system starts with SElLinux enabled in a permissive mode.

Again, I'm not sure if disabling SELinux should mean enabling AppArmor. I don't think so.

dgdavid: Ok, I have digging a little bit more on the _major security LSM module_ and my conclusions are

* Unless I'm wrong, AppArmor is already enabled since kernel was compiled with [`CONFIG_DEFAULT_SECURITY_APPARMOR=y`](https://github.com/torvalds/linux/commit/2623c4fbe2ad1341ff2d1e12410d0afdae2490ca) and `CONFIG_LSM="integrity,apparmor"`, as it can be seen in the _/boot/config-\<kernel-version\>_ file (at least in openSUSE using a 5.10 kernel and SLES using a 5.3)
* [According to the documentation](https://www.kernel.org/doc/html/v5.6/admin-guide/LSM/apparmor.html#how-to-enable-disable), AppArmor will be disabled when using `apparmor=0` or `security=xxx` at boot time
* So, _deleting_ the `security=selinux` boot param looks safe since the user has to disable AppArmor on purpuose (either, with the `apparmor` param or using another major security module, like SELinux)

I did some tests and took the screenshots you can see below

| Not using `security` | `security=selinux` and `apparmor=1` | `security=apparmor` and `selinux=1`|
| - | - | - |
| no_params | security_selinux_and_apparmor_1 | security_apparmor_and_selinux_1 |


Maybe we could have a more generic `LSMSecurityConfig` or `SecurityConfig` class to manage the security module to use (`security=xxx xxx=1`) and ask to its backend class for its configuration/policies/whatever. But I think it is out of the scope of this task.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    def initialize
+      propose_policy if Yast::Mode.installation
+      @policy = configured_policy
+    end
+
+    # Returns the policy applied in the running system
+    def running_policy
+      Yast::Execute.locally!(GETENFORCE_PATH, stdout: :capture).chomp.downcase.to_sym
+    rescue Cheetah::ExecutionFailed
```
jreidinger, line 39: this can actually hide other error states. So I would not use debug unless exit code is 127

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,118 @@
...
+    # Use the given policy for the next boot (in running system is needed to #save)
+    def update_policy(policy_key)
+      found_policy = find_policy(policy_key)
+
+      if found_policy
+        log.info("Changing SELinux configuration to `#{policy_key}` policy: #{found_policy}")
+
+        @policy = policy_key
+      else
+        log.info("Unknown `#{policy_key}` SELinux policy")
```
jreidinger, line 58: I would use at least warning here. Maybe even exception is better.

src/lib/security/selinux_config.rb
dgdavid: >  So I would not use debug unless exit code is 127

Did you mean _:disabled_ instead _debug_?

jreidinger: I mean that you can consider missing tool as :disabled, but you rescue all error states including e.g. 1 or 2 which can mean something different

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,182 @@
...
+      log.debug("`#{GETENFORCE_PATH}` not available. SELinux or selinux-tools missing")
+      :disabled
+    end
+
+    # Set the policy to given value
+    #
+    # @note using nil means to set SELinux policy as disabled.
+    #
+    # @param id [String, Symbol, nil] a SELinux policy identifier
+    # @return [] true when policy found and set; false otherwise
```
dgdavid, line 69: This is a leftover, I have to change it

src/lib/security/selinux_config.rb
```diff
@@ -25,92 +25,156 @@ module Security
...
+
+    # Return the initial SELinux policy
+    #
+    # It can be the proposed one if running in an installation or the configured policy for a
+    # running system
+    #
+    # @return [Symbol]
+    def initial_policy
+      @initial_policy ||=
+        begin
```
jreidinger, line 45: I would keep it more simple like `return @initial_policy if @initial_policy`

src/lib/security/selinux_config.rb
```diff
@@ -25,92 +25,156 @@ module Security
...
       log.debug("`#{GETENFORCE_PATH}` not available. SELinux or selinux-tools missing")
       :disabled
     end
 
-    # If current policy value is different to the configured one
-    def changed?
-      @policy != configured_policy
+    # Set the policy to given value
+    #
+    # @note using nil means to set SELinux policy as disabled.
```
jreidinger, line 66: why the nil case is needed?

src/lib/security/selinux_config.rb
```diff
@@ -25,92 +25,156 @@ module Security
...
-
+    # Set current policy options as kernel parameters for the next boot
+    #
+    # @note it does not write the changes when running in installation mode, where only sets the
+    #   kernel params in memory, since the Yast::Bootloader.Write will be performed at the end of
+    #   installation.
+    #
+    # @see Yast::Bootloader#modify_kernel_params
+    #
+    # @return [Boolean] false if there area not policy options or nothing changed; true otherwise
```
jreidinger, line 82: typo area. And I also do not like much case that wrong policy and not changed policy has same return state.

src/lib/security/selinux_config.rb
```diff
@@ -25,92 +25,156 @@ module Security
...
+    #
+    # @return [Symbol] the policy identifier
+    def configured_policy
+      options = policy_from_kernel_params
+
+      return :disabled if options.empty?
+
+      security_module = options["security"]
+      module_disabled = options["selinux"].to_i <= 0
+      enforcing_mode  = options["enforcing"].to_i > 0
```
jreidinger, line 159: so e.g. enforcing=15 is valid?

src/lib/security/selinux_config.rb
dgdavid: Actually not needed, it is just a _facility_ for disabling it. We can remove it.

src/lib/security/selinux_config.rb
```diff
@@ -25,92 +25,156 @@ module Security
...
     end
 
-    def policy_options_from_kernel
-      Hash[*POLICY_KEYS.flat_map { |key| [key, read_param(key)] }]
+    # Returns the SELinux configuration based on options set in the kernel command line
+    #
+    # @return [Symbol] the policy identifier
+    def policy_from_kernel_params
+      options = Hash[*POLICY_KEYS.flat_map { |key| [key, read_param(key)] }]
+      options.filter { |_, value| value != :missing }
```
jreidinger, line 172: still this code does not address if there are multiple entries on cmdline. E.g. consider `security=selinux selinux=0 enforcing=0 selinux=1 enforcing=1`

src/lib/security/selinux_config.rb
dgdavid: Yes, it is.

Screenshot_opensusetumbleweed_2021-02-02_10:36:06

src/lib/security/selinux_config.rb
dgdavid: Right.

But It's a little bit crazy to know how to proceed in such cases. See my tests below

#### `security=selinux selinux=0 enforcing=0 selinux=1 enforcing=1`

SELinux Permissive wins

```
localhost:~ # cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-5.10.1-1-default root=UUID=1d930088-0e08-4013-b2fe-5e639eddf91a splash=silent quiet security=selinux selinux=0 enforcing=0 selinux=1 enforcing=1
localhost:~ # getenforce
Permissive
```

#### `security=selinux selinux=0 enforcing=1 selinux=1 enforcing=0`

SELinux Enforcing wins

```
localhost:~ # cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-5.10.1-1-default root=UUID=1d930088-0e08-4013-b2fe-5e639eddf91a splash=silent quiet security=selinux selinux=0 enforcing=1 selinux=1 enforcing=0
localhost:~ # getenforce
Enforcing
```


#### `security=selinux selinux=1 enforcing=0 selinux=0 enforcing=1`

SELinux Disabled wins

```
localhost:~ # cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-5.10.1-1-default root=UUID=1d930088-0e08-4013-b2fe-5e639eddf91a splash=silent quiet security=selinux selinux=1 enforcing=0 selinux=0 enforcing=1
localhost:~ # getenforce
Disabled
```

jreidinger: yeah, that looks really crazy..hopefully bootloader API returns array in order, so it is solvable to implement such logic, I am just not sure how much backward compatible it is Probably the most intersting question for runtime module.

dgdavid: > hopefully bootloader API returns array in order

That's what I thought and what I was testing, but seems that it returns the array in the same order,  not in the _right_ one. See attached screenshots

Screenshot_opensusetumbleweed_2021-02-02_12:15:37

Screenshot_opensusetumbleweed_2021-02-02_12:19:12

Screenshot_opensusetumbleweed_2021-02-02_12:24:31

jreidinger: yes, it does not have such knowledge, so it returns order as it is on cmdline. It is up to you to decide which value wins or if you need to merge it ( as e.g. console means that each value is another device to send output ).

jreidinger: My main objection here is that you use `to_i` on result of that get_param which crash if there are more keys

dgdavid: Ok, as we agreed in a video call, lets go for the latest value for now.

Thanks Josef.

dgdavid:
As stated in the description, this was a draft PR  in the context of [1] for discussing about modeling a class to manage SELinux configuration.

So, let's close it now in favor of a new one to keep the review less noisy.

Thanks

[1] https://trello.com/c/g94xQBDM/2159-8-musthave-selinux-and-security-refac (internal link)

src/lib/security/selinux_config.rb
dgdavid: Finally I decided to simple log the exception message, see https://github.com/yast/yast-security/pull/83/files#diff-0a09e81c830190bb29f23761529237b497379394ad66db6a78b8d671f49791c3R63-R70. We can change it later if we still thinking it is really needed.

PR #83, opened 2021-02-02
Add class to manage SELinux configuration

dgdavid:
## Feature

To have a back-end class for managing the SELinux configuration at boot time

### About completely removing `security` param when disabling SELinux

> * AppArmor is already enabled since kernel was compiled with [`CONFIG_DEFAULT_SECURITY_APPARMOR=y`](https://github.com/torvalds/linux/commit/2623c4fbe2ad1341ff2d1e12410d0afdae2490ca) and `CONFIG_LSM="integrity,apparmor"`, as it can be seen in the _/boot/config-\<kernel-version\>_ file (at least in openSUSE using a 5.10 kernel and SLES using a 5.3)
> * [According to the documentation](https://www.kernel.org/doc/html/v5.6/admin-guide/LSM/apparmor.html#how-to-enable-disable), AppArmor will be disabled when using `apparmor=0` or `security=xxx` at boot time
> * So, **_deleting_ the `security=selinux` boot param looks safe** since the user has to disable AppArmor on purpose (either, with the `apparmor` param or using another major security module, like SELinux)

<sub>Extracted from https://github.com/yast/yast-security/pull/82#discussion_r567817487</sub>

### About the `security` parameter

Although the feature request clearly states using `security=`, in the [kernel admin-guide](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html) is marked as deprecated in favor of `lsm=`

>   security=       [SECURITY] Choose a legacy "major" security module to enable at boot. This has been deprecated by the "lsm=" parameter.

## Tests

* Added unit tests
* Tested manually

  <details>
    <summary>Click to show/hide some screenshots</summary>

   ---

   | Working with `SelinuxConfig` in an `irb` console | System after using `SelinuxConfig` for switching to permissive mode |
   |-|-|
   | selinux_config_in_irb_session | system_switched_to_permissive_selinux_mode |

  </details>

## Still pending

* Learn how selinux kernel params _precedence_ works. See https://github.com/yast/yast-security/pull/82#discussion_r568512761 for more details.

---

<sub>This can be considered the result of #82</sub>
<sub>Related to https://trello.com/c/g94xQBDM/2159-8-musthave-selinux-and-security-refac (internal link)</sub>

### DIFF
--- a/src/lib/security/selinux_config.rb
+++ b/src/lib/security/selinux_config.rb
@@ -22,6 +22,34 @@ require "yast2/execute"
 
 module Security
   # Class for handling SELinux kernel params
+  #
+  # @example Querying the currently configured SELinux mode
+  #   selinux = SelinuxConfig.new
+  #   mode = selinux.mode
+  #   puts mode.id #=> :permissive
+  #   puts mode.name #=> "Permisive"
+  #
+  # @example Querying the currently running SELinux mode
+  #   selinux= SelinuxConfig.new
+  #   mode = selinux.running_mode
+  #   puts mode.id #=> :enforcing
+  #   puts mode.name #=> "Enforcing"
+  #
+  # @example Enabling SELinux in Permissive mode for next boot
+  #   selinux = SelinuxConfig.new
+  #   selinux.mode = :permissive
+  #   selinux.save #=> true
+  #
+  # @example Disabling SELinux for next boot
+  #   selinux = SelinuxConfig.new
+  #   selinux.mode = :disabled
+  #   selinux.save #=> true
+  #
+  # @example Disabling SELinux for next boot (using nil)
+  #   selinux = SelinuxConfig.new
+  #   selinux.mode = nil
+  #   selinux.mode.id  #=> :disabled
+  #   selinux.save #=> true
   class SelinuxConfig
     extend Yast::I18n
     include Yast::I18n
--- a/src/lib/security/selinux_config.rb
+++ b/src/lib/security/selinux_config.rb
@@ -1,0 +1,263 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "yast2/execute"
+
+module Security
+  # Class for handling SELinux kernel params
+  class SelinuxConfig
+    extend Yast::I18n
+    include Yast::I18n
+    include Yast::Logger
+
+    Yast.import "Bootloader"
+    Yast.import "ProductFeatures"
+
+    # @return [Security::SelinuxConfig::Mode] the last set mode, which can be differrent to the
+    #   {#running_mode} and {#configured_mode}. A call to {#save} is needed to make it the
+    #   {#configured_mode} for the next boot.
+    attr_reader :mode
+
+    # Constructor
+    def initialize
+      @mode = initial_mode
+    end
+
+    # Returns the initial SELinux mode
+    #
+    # It can be the proposed one if running in an installation or the configured mode for a
+    # running system
+    #
+    # @return [Security::SelinuxConfig::Mode]
+    def initial_mode
+      return @initial_mode if @initial_mode
+
+      propose! if Yast::Mode.installation
+
+      @initial_mode = configured_mode
+    end
+
+    # Returns the mode applied in the running system
+    #
+    # @note the system can be booted with a different SELinux mode that the configured one. To
+    #  know the running mode getenforce tool is used.
+    #
+    # @return [Mode] running SELinux mode if command executed successfully; :disabled otherwise
+    def running_mode
+      id = Yast::Execute.locally!(GETENFORCE_PATH, stdout: :capture).chomp.downcase.to_sym
+      Mode.find(id)
+    rescue Cheetah::ExecutionFailed => e
+      log.info(e.message)
+
+      Mode.find(:disabled)
+    end
+
+    # Returns a collection holding all known SELinux modes
+    #
+    # @return [Array<Security::SelinuxConfig::Mode>] a collection of knwon SELinux modes
+    def modes
+      Mode.all
+    end
+
+    # Set the mode to given value
+    #
+    # @note using nil means to set SELinux mode as disabled.
+    #
+    # @param id [String, Symbol, nil] a SELinux mode identifier
+    # @return [Mode, nil] found SelinuxConfig::Mode by given id; nil if none
+    def mode=(id)
+      @mode = Mode.find(id)
+    end
+
+    # Set current mode options as kernel parameters for the next boot
+    #
+    # @note it does not write the changes when running in installation mode, where only sets the
+    #   kernel params in memory, since the Yast::Bootloader.Write will be performed at the end of
+    #   installation.
+    #
+    # @see Yast::Bootloader#modify_kernel_params
+    #
+    # @return [Boolean] false if mode is not set or nothing changed; true otherwise
+    def save
+      return false if mode.nil?
+
+      changed = Yast::Bootloader.modify_kernel_params(**mode.options)
+      changed = Yast::Bootloader.Write if changed && !Yast::Mode.installation
+      changed
+    end
+
+    private
+
+    # Path to the SELinux getenforce command
+    GETENFORCE_PATH = "/usr/sbin/getenforce".freeze
+    private_constant :GETENFORCE_PATH
+
+    # Proposes a mode based on `selinux_mode` value set in the control file
+    #
+    # If mode is found, it calls #save for setting it.
+    #
+    # @return [Mode] the configured SELinux mode after trying the proposal
+    def propose!
+      id = Yast::ProductFeatures.GetFeature("globals", "selinux_mode").to_sym
+
+      if Mode.find(id)
+        log.info "Proposing the `#{id}` SELinux mode"
+        self.mode = id
+        save
+      else
+        log.info "Unknown `#{id}` SELinux mode. Skipping the proposal."
+      end
+
+      self.mode
+    end
+
+    # Returns the configured SELinux mode according to params in kernel command line
+    #
+    # @see #mode_from_kernel_params
+    #
+    # @return [Symbol] the mode identifier
+    def configured_mode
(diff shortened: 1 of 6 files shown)


src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+    # Returns the initial SELinux mode
+    #
+    # It can be the proposed one if running in an installation or the configured mode for a
+    # running system
+    #
+    # @return [Security::SelinuxConfig::Mode]
+    def initial_mode
+      return @initial_mode if @initial_mode
+
+      propose! if Yast::Mode.installation
```
jreidinger, line 52: result is not stored anywhere. I expect you want something like `@initial_mode = Yast::Mode.installation ? propose! : configured_mode`
and that reminds me why we need propose with side-efect? I do not like it much. Why not havve propose that just return proposed mode?

jreidinger: maybe to be more consistent I would use `proposed_mode` as method name and result can be also nil if nothing should be proposed? not sure

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+      Mode.find(id)
+    rescue Cheetah::ExecutionFailed => e
+      log.info(e.message)
+
+      Mode.find(:disabled)
+    end
+
+    # Returns a collection holding all known SELinux modes
+    #
+    # @return [Array<Security::SelinuxConfig::Mode>] a collection of knwon SELinux modes
```
jreidinger, line 74: typo knwon

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+    def modes
+      Mode.all
+    end
+
+    # Set the mode to given value
+    #
+    # @note using nil means to set SELinux mode as disabled.
+    #
+    # @param id [String, Symbol, nil] a SELinux mode identifier
```
jreidinger, line 83: I think it makes sense to also accept Mode instance. I also missing that nil handling below ( maybe I just overlook it ).

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+    # @return [Mode, nil] found SelinuxConfig::Mode by given id; nil if none
```
jreidinger, line 84: assignment does not have return.

lets try it youself:
```
jreidinger@50-3E-AA-D7-45-9F:~/prace/yast/security/src> irb
irb(main):001:0> def a=(test);false;end
=> :a=
irb(main):002:0> self.a = 15
=> 15
```

jreidinger: it always return parameter

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+
+    # Set current mode options as kernel parameters for the next boot
+    #
+    # @note it does not write the changes when running in installation mode, where only sets the
+    #   kernel params in memory, since the Yast::Bootloader.Write will be performed at the end of
+    #   installation.
+    #
+    # @see Yast::Bootloader#modify_kernel_params
+    #
+    # @return [Boolean] false if mode is not set or nothing changed; true otherwise
```
jreidinger, line 97: to be honest I do not like that false for nothing change. If nothing change, so it is already properly set, then I prefer true. So false is really kept only for failed write, so false response is always problem.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+    def save
+      return false if mode.nil?
+
+      changed = Yast::Bootloader.modify_kernel_params(**mode.options)
+      changed = Yast::Bootloader.Write if changed && !Yast::Mode.installation
```
jreidinger, line 102: you can write it as `changed &&= Yast::Bootloader.Write unless Yast::Mode.installation`

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+
+    private
+
+    # Path to the SELinux getenforce command
+    GETENFORCE_PATH = "/usr/sbin/getenforce".freeze
+    private_constant :GETENFORCE_PATH
+
+    # Proposes a mode based on `selinux_mode` value set in the control file
+    #
+    # If mode is found, it calls #save for setting it.
```
jreidinger, line 114: as said above I do not like methods with side effects and that save is really strange for me.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+      #
+      # @param id [String, Symbol] if of the mode
+      # @param name [String] the mode name, a string marked for translation
+      # @param enable [Boolean] wether the mode will enable SELinux
+      # @param enforcing [Boolean] if SELinux should be run enforcing or not
+      def initialize(id, name, enable, enforcing)
+        textdomain "security"
+
+        @id = id.to_sym
+        @name = _(name)
```
jreidinger, line 226: I worry this cannot be done. As during proposal user can change language then we need also strings retranslated. So instead of hardcoded translation I would store here marked name and have method
```
def to_human_string
  _(@name)
end
```

This way it always return string respecting *current* language. Common pitfall in installation when language can change easily.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+
+      # Returns all known keys for setting a SELinux mode via kernel command line
+      #
+      # @return [Array<String>]
+      def self.keys
+        OPTIONS_KEYS
+      end
+
+      # Finds a SELinux mode by its id
+      #
```
jreidinger, line 192: missing param documentation. I expect it should work with String,Symbol or Mode. But with nil it crash.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+        find(:enforcing)
+      end
+
+      # Constructor
+      #
+      # Intended to be used internally by the class
+      #
+      # @param id [String, Symbol] if of the mode
+      # @param name [String] the mode name, a string marked for translation
+      # @param enable [Boolean] wether the mode will enable SELinux
```
jreidinger, line 220: type wether

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+      #
+      # @return [Array<Mode>]
+      def self.all
+        ALL.dup
+      end
+
+      # Returns all known keys for setting a SELinux mode via kernel command line
+      #
+      # @return [Array<String>]
+      def self.keys
```
jreidinger, line 187: I prefer better method name like `kernel_options` or something like it. Otherwise it looks like some mode keys, which is a bit strange.

src/lib/security/selinux_config.rb
```diff
@@ -0,0 +1,263 @@
...
+      alias_method :to_human_string, :name
+
+      # @return [Hash{String=><String, :missing>}] options for setting the mode via kernel params
+      attr_reader :options
+
+      # Returns all known SELinux modes
+      #
+      # @return [Array<Mode>]
+      def self.all
+        ALL.dup
```
jreidinger, line 181: is modificable list needed?

src/lib/security/selinux_config.rb
dgdavid: I thought that we needed to _store_ it as boot params to make the proposal effective. But, according to your concern, seem like it is not necessary. Honestly, I don't like side-effects neither. So, I'll happyly to getride of it and adapt this method too :)

src/lib/security/selinux_config.rb
dgdavid: Yes, it make sense. Actually, I added the :to_sym alias method in`Mode` to accept it, but then I forget to update the documentation :)

src/lib/security/selinux_config.rb
dgdavid: Arrrrg!!! Stupid mistake

src/lib/security/selinux_config.rb
dgdavid: The `nil` thingy is actually a leftover from previous code. But I'll add the handling to return a disabled mode in such case.

src/lib/security/selinux_config.rb
dgdavid: Good to know. I didn't realize that it is _kind of_ hard-coded translation. Thanks

src/lib/security/selinux_config.rb
jreidinger: well, if we need it, we will use the save on place where it is needed. it will be move obvious.

src/lib/security/selinux_config.rb
jreidinger: well, it works that `_()` return localized string and you remember it. So if later proposal change language it is still memorized.

src/lib/y2security/selinux_config.rb
```diff
@@ -0,0 +1,290 @@
...
+  #   selinux = SelinuxConfig.new
+  #   selinux.mode = :permissive
+  #   selinux.save #=> true
+  #
+  # @example Disabling SELinux for next boot
+  #   selinux = SelinuxConfig.new
+  #   selinux.mode = :disabled
+  #   selinux.save #=> true
+  #
+  # @example Disabling SELinux for next boot (using nil)
```
jreidinger, line 48: is there way to simply do nothing? I would like to use it when control file say do not touch it, then I would like to hide UI, but in backend just set it as keep it.

src/lib/y2security/selinux_config.rb
```diff
@@ -0,0 +1,290 @@
...
+    def mode=(id)
+      found_mode = Mode.find(id)
+
+      if found_mode.nil?
+        log.info("Requested SELinux mode `#{id}` not found. Falling back to :disabled.")
+        found_mode = Mode.find(:disabled)
+      end
+
+      @mode = found_mode
+      @mode
```
jreidinger, line 116: this line is not needed.

src/lib/y2security/selinux_config.rb
```diff
@@ -0,0 +1,290 @@
...
+    # Path to the SELinux getenforce command
+    GETENFORCE_PATH = "/usr/sbin/getenforce".freeze
+    private_constant :GETENFORCE_PATH
+
+    # Proposes a mode based on `selinux_mode` value set in the control file
+    #
+    # @see Mode.find
+    #
+    # @return [Mode] disabled or found SELinux mode
+    def propose_mode
```
jreidinger, line 147: I would call it `proposed_mode` to be consistent with configured mode and also makes more sense.

src/lib/y2security/selinux_config.rb
dgdavid: Another lerftover from previous code base, sorry.

src/lib/y2security/selinux_config.rb
dgdavid: @jreidinger see https://github.com/yast/yast-security/pull/83/commits/2b58a20cbe09fb17ea1b4ddf1ac61fd3cf44fded

We can revert it if it is not what you were looking for :)

dgdavid: Finally reverted in favor of [`#configurable?`](https://github.com/yast/yast-security/pull/83/commits/2ccd5c23b7bc1f6fd77d6b9e8ff55d199a1b6258)

jreidinger (approved):

PR #85, opened 2021-02-05
WIP: AY for SELinux

schubi2:


src/modules/Security.rb
```diff
@@ -867,6 +901,21 @@ def default_encrypt_method
 
     protected
 
+    SELINUX_PACKAGE = "selinux-policy-targeted"
```
schubi2, line 904: That is open. E.G. it could comes from the product control file.

dgdavid: Now we have the [`Y2Security::Selinux#needed_patterns`](https://github.com/yast/yast-security/blob/7c23e55f7af44eae59691c91ef5a65c70c0b497e/src/lib/y2security/selinux.rb#L199-L205) method. Not sure if useful here.

dgdavid:
Continues in #88

PR #86, opened 2021-02-08
Add SelinuxConfig#needed_packages method

dgdavid:


src/lib/y2security/selinux_config.rb
```diff
@@ -139,12 +140,34 @@ def configurable?
       Yast::ProductFeatures.GetBooleanFeature("globals", "selinux_configurable")
     end
 
+    def needed_packages
+      pattern = Yast::OSRelease.ReleaseName =~ /micro/i ? MICRO_OS_PATTERN : PATTERN
+
+      [pattern, TOOLS_PACKAGE, POLICY_PACKAGE]
+    end
+
```
dgdavid, line 148: Not sure if this is the right way to proceed or it would be better asking to the package system or even other way. What do you think?

src/lib/y2security/selinux_config.rb
```diff
@@ -139,12 +140,34 @@ def configurable?
...
+    # SELinux tools package
+    TOOLS_PACKAGE = "selinux-tools".freeze
+    private_constant :TOOLS_PACKAGE
+
+    # SELinux pattern
+    PATTERN = "selinux".freeze
+    private_constant :PATTERN
+
+    # SLE Micro OS pattern
+    MICRO_OS_PATTERN = "microos-selinux".freeze
```
dgdavid, line 168: Based on https://github.com/yast/skelcd-control-SMO/blob/SLE-15-SP2/control/control.SMO.xml#L89

dgdavid:
Go to #87 instead.

PR #87, opened 2021-02-11
Refactor Y2Security::SelinuxConfig (now Y2Security::Selinux)

dgdavid:
Related to [1] and [2], this PR introduces changes and improvements in the former `Y2Security::SelinuxConfig` class added in #83 and now renamed to `Y2Security::Selinux`. As consequence of its improvements, a new `CFA::Selinux` class has been added to handle the SELinux configuration file `/etc/selinux/config` (only able to read and write the `SELINUX` value by now)

Some of the changes that worth to mention are

* The _#mode_ proposal has become _lazy_, instead of triggering it in the constructor (agreed after working on https://github.com/yast/yast-firewall/pull/144, where too much mock had to be used to fix the test suite, which was a kind of symptom that the proposal was _miss-placed_)
* Improved the _#mode_ loading/detection, which is not as easy as it looks at glimpse because of _precedence_ and _validness_ of the SELinux boot params (see [_#boot\_mode_](https://github.com/yast/yast-security/blob/ee5019c74ad0d24713f78be5c504c6c99cc10927/src/lib/y2security/selinux.rb#L124-L146), the former _#configured\_mode_)
* Fixed the _#configured\_mode_, which is now the one defined as a `SELINUX` value in the configuration file.
* Added `CFA::Selinux` class to handle the SELinux configuration file (/etc/selinux/config).
* Added a method `#needed_patterns`, useful to know which patterns are required/recommended to has SELinux workins as expected (see https://github.com/yast/yast-installation-control/pull/108). NOTE: as @lslezak said in the `#yast` IRC channel at the `freenode.net` server, _PackagesProposal (despite the name) also supports patterns_
* Adapted/improvied unit testing

---

:warning: **NOTE**: `semanage` lens [has been replaced by `simplevars`](https://github.com/yast/yast-security/pull/87/commits/f7b428696fe33814530c243511d52c42822c7587) because in SP2 products the available augeas-lenses is 1.10.1 and [`semanage` was introduced in augeas-lenses 1.12](https://github.com/hercules-team/augeas/pull/594/files#diff-7ee66c4f1536ac84dc5bbff1b8312e2eef24b974b3e48a5c5c2bcfdf2eb8f3ceR5-R6).

---

[1] https://trello.com/c/g94xQBDM/2159-8-musthave-selinux-and-security-refac (internal link)
[2] https://jira.suse.com/browse/SLE-17342 (internal link)

---
<sub>Additionally, some JIRA reference has been fixed in the changelog file</sub>

### DIFF
--- a/package/yast2-security.spec
+++ b/package/yast2-security.spec
@@ -32,7 +32,7 @@ BuildRequires:  perl-XML-Writer
 BuildRequires:  update-desktop-files
 BuildRequires:  yast2-pam
 BuildRequires:  yast2-devtools >= 4.2.2
-# Y2Security::SelinuxConfig requires Yast::Bootloader
+# Y2Security::Selinux requires Yast::Bootloader
 BuildRequires:  yast2-bootloader
 BuildRequires:  rubygem(%{rb_default_ruby_abi}:yast-rake) >= 0.2.5
 BuildRequires:  rubygem(%{rb_default_ruby_abi}:rspec)
@@ -50,8 +50,8 @@ Requires:       yast2-pam >= 2.14.0
 # CFA::SysctlConfig
 Requires:       yast2 >= 4.2.66
 Requires:       yast2-ruby-bindings >= 1.0.0
-# Y2Security::SelinuxConfig requires Yast::Bootloader
-BuildRequires:  yast2-bootloader
+# Y2Security::Selinux requires Yast::Bootloader
+Requires:       yast2-bootloader
 
 Provides:       y2c_sec yast2-config-security
 Provides:       yast2-trans-security y2t_sec
--- a/package/yast2-security.spec
+++ b/package/yast2-security.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-security
-Version:        4.2.15
+Version:        4.2.16
 Release:        0
 Group:          System/YaST
 License:        GPL-2.0-only
--- a/package/yast2-security.spec
+++ b/package/yast2-security.spec
@@ -38,6 +38,8 @@ BuildRequires:  rubygem(%{rb_default_ruby_abi}:yast-rake) >= 0.2.5
 BuildRequires:  rubygem(%{rb_default_ruby_abi}:rspec)
 # CFA::SysctlConfig
 BuildRequires:  yast2 >= 4.2.66
+# CFA::Selinux
+BuildRequires:  augeas-lenses
 # Unfortunately we cannot move this to macros.yast,
 # bcond within macros are ignored by osc/OBS.
 %bcond_with yast_run_ci_tests
@@ -52,6 +54,8 @@ Requires:       yast2 >= 4.2.66
 Requires:       yast2-ruby-bindings >= 1.0.0
 # Y2Security::Selinux requires Yast::Bootloader
 Requires:       yast2-bootloader
+# CFA::Selinux
+Requires:       augeas-lenses
 
 Provides:       y2c_sec yast2-config-security
 Provides:       yast2-trans-security y2t_sec
--- a/src/lib/y2security/selinux.rb
+++ b/src/lib/y2security/selinux.rb
@@ -240,9 +240,7 @@ module Y2Security
     #
     # @return [CFA::Selinux]
     def config_file
-      return @config_file if @config_file
-
-      @config_file = CFA::Selinux.load
+      @config_file ||= CFA::Selinux.load
     end
 
     # Sets the mode to the proposed one via `selinux_mode` global variable in the control file
--- a/src/lib/y2security/selinux.rb
+++ b/src/lib/y2security/selinux.rb
@@ -19,6 +19,7 @@
 
 require "yast"
 require "yast2/execute"
+require "cfa/selinux"
 
 module Y2Security
   # Class for handling SELinux kernel params
@@ -73,6 +74,13 @@ module Y2Security
       @mode ||= make_proposal || configured_mode
     end
 
+    # Returns the configured mode in the SELinux config file
+    #
+    # @return [Mode, nil] the SELinux mode set in the config file; nil if unknown or not set
+    def configured_mode
+      Mode.find(config_file.selinux)
+    end
+
     # Returns the mode applied in the running system
     #
     # @note the system can be booted with a different SELinux mode that the configured one. To
@@ -118,6 +126,8 @@ module Y2Security
       return false unless configurable?
 
       Yast::Bootloader.modify_kernel_params(mode.options)
+      config_file.selinux = mode.id.to_s
+      config_file.save
 
       return true if Yast::Mode.installation
 
@@ -163,6 +173,15 @@ module Y2Security
       @product_feature_settings = settings
     end
 
+    # Returns a CFA::Selinux object for handling the config file
+    #
+    # @return [CFA::Selinux]
+    def config_file
+      return @config_file if @config_file
+
+      @config_file = CFA::Selinux.load
+    end
+
     # Find SELinux mode by given value
     #
     # @note using nil means to set SELinux mode as disabled.
@@ -202,15 +221,6 @@ module Y2Security
       log.info("Proposing `#{id}` SELinux mode.")
 
       find_mode(id)
-    end
-
-    # Returns the configured SELinux mode according to params in kernel command line
-    #
-    # @see #mode_from_kernel_params
-    #
-    # @return [Symbol] the mode identifier
-    def configured_mode
-      Mode.match(mode_from_kernel_params)
     end
 
     # Returns the SELinux configuration based on options set in the kernel command line
--- a/src/lib/y2security/selinux.rb
+++ b/src/lib/y2security/selinux.rb
@@ -124,6 +124,13 @@ module Y2Security
       Yast::Bootloader.Write
     end
 
+    # Returns needed patterns defined in the product features
+    #
+    # @return [String] defined patterns in product features to have SELinux working as expected
+    def needed_patterns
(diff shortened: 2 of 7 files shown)


src/lib/y2security/selinux.rb
```diff
@@ -122,21 +172,30 @@ def save
       return false unless configurable?
 
       Yast::Bootloader.modify_kernel_params(mode.options)
```
dgdavid, line 174: Just a note, perhaps the calls to the `Bootloader` calls should be made elsewhere, in the _client_ using it and this method be only responsible to update the needed values in the SELinux configuration file (CFA::Selinux, /etc/selinux/config)

package/yast2-security.spec
```diff
@@ -32,12 +32,14 @@ BuildRequires:  perl-XML-Writer
...
 BuildRequires:  yast2-devtools >= 4.2.2
-# Y2Security::SelinuxConfig requires Yast::Bootloader
+# Y2Security::Selinux requires Yast::Bootloader
 BuildRequires:  yast2-bootloader
 BuildRequires:  rubygem(%{rb_default_ruby_abi}:yast-rake) >= 0.2.5
 BuildRequires:  rubygem(%{rb_default_ruby_abi}:rspec)
 # CFA::SysctlConfig
 BuildRequires:  yast2 >= 4.2.66
+# CFA::Selinux
+BuildRequires:  augeas-lenses
```
dgdavid, line 42: Realizing now that it is actually not needed because of the `yast2 >= 4.2.66` requirement (https://github.com/yast/yast-yast2/blob/64ff8b58e57ae84448f099ff4f97b0d92b76bfeb/package/yast2.spec), although it does not hurt.

jreidinger: yep, better specify it in each package so we do not depend on others and tight with them

src/lib/y2security/selinux.rb
```diff
@@ -19,101 +19,167 @@
...
+    #
+    # @param id [Selinux::Mode, String, Symbol, nil] a SELinux mode or its identifier
+    # @return [Mode] the Selinux::Mode by given id or disabled is none found or nil was given
     def mode=(id)
-      @mode = find_mode(id)
-      @mode
+      found_mode = Mode.find(id)
+
+      if found_mode.nil?
+        log.info("Requested SELinux mode `#{id}` not found. Falling back to :disabled.")
```
jreidinger, line 165: I would use here warn or err instead

src/lib/y2security/selinux.rb
schubi2: I think it is not so important anymore to do this because you have renamed the class to selinux and that could include the bootloader call.

src/lib/y2security/selinux.rb
```diff
@@ -122,21 +188,30 @@ def save
...
 
       return true if Yast::Mode.installation
 
       Yast::Bootloader.Write
     end
 
+    # Returns needed patterns defined in the product features
+    #
+    # @return [String] defined patterns in product features to have SELinux working as expected
+    def needed_patterns
```
schubi2, line 202: What does it return if there is no definition ?

src/lib/y2security/selinux.rb
```diff
@@ -168,20 +261,17 @@ def find_mode(id)
...
+      id = product_feature_settings[:mode]
+      found_mode = Mode.find(id)
 
-      log.info("Proposing `#{id}` SELinux mode.")
+      if found_mode.nil?
+        log.info("Proposed SELinux mode `#{id}` not found.")
 
-      find_mode(id)
-    end
+        return
```
schubi2, line 270: Is there the return value missed ? Or do I have missed something ? :-)

src/lib/y2security/selinux.rb
```diff
@@ -122,21 +188,30 @@ def save
...
+      config_file.save
 
       return true if Yast::Mode.installation
 
       Yast::Bootloader.Write
     end
 
+    # Returns needed patterns defined in the product features
+    #
+    # @return [String] defined patterns in product features to have SELinux working as expected
```
jreidinger, line 201: I expect here array as it is `patterns` and also if it is separated by space it should be detail.

src/lib/y2security/selinux.rb
dgdavid: It returns and empty string, if I'm not wrong

src/lib/y2security/selinux.rb
```diff
@@ -145,21 +220,39 @@ def configurable?
...
     #
-    # @note using nil means to set SELinux mode as disabled.
+    # @return [Hash{Symbol => String, Boolean, nil}] e.g., { mode: "enforcing", configurable: false }
+    #   a hash holding the defined SELinux options in the control file;
+    #   an empty object if no settings are defined
+    def product_feature_settings
+      @product_feature_settings unless @product_feature_settings.nil?
+
+      settings = Yast::ProductFeatures.GetFeature("globals", "selinux")
+      settings = {} if settings.empty?
```
jreidinger, line 232: so it can be empty string? I do not get this line

src/lib/y2security/selinux.rb
dgdavid: Ok, will change.

src/lib/y2security/selinux.rb
dgdavid: If I'm not wrong, when it is not defined ProductFeatures returns and empty string. See https://github.com/yast/yast-yast2/blob/e914214656d217ae05a51f6347e22b520caae907/library/control/src/modules/ProductFeatures.rb#L203-L213

src/lib/y2security/selinux.rb
```diff
@@ -145,21 +220,39 @@ def configurable?
...
-      found_mode = Mode.find(id)
+    # @return [CFA::Selinux]
+    def config_file
+      return @config_file if @config_file
 
-      if found_mode.nil?
-        log.info("Requested SELinux mode `#{id}` not found. Falling back to :disabled.")
-        found_mode = Mode.find(:disabled)
-      end
+      @config_file = CFA::Selinux.load
```
jreidinger, line 244: can be simple `@config_file ||= CFA::Selinux.load`

jreidinger: also I think exception should be catched if file does not exists?

src/lib/y2security/selinux.rb
jreidinger, line 270: empty return? strange I do not expect nil there

src/lib/y2security/selinux.rb
dgdavid: Right, that method was written before adding the class method `.load` and then it was not fully adapted. Good catch, thanks!

src/lib/y2security/selinux.rb
dgdavid: It has to return nil if the mode was not found. That's because we are using it as the first option (lazy proposal) for the first call to `#mode`

dgdavid: Not missing, returning `nil`

src/lib/y2security/selinux.rb
dgdavid: Now it returns an empty array. See https://github.com/yast/yast-security/pull/87#discussion_r574408651 and https://github.com/yast/yast-security/pull/87/commits/9191e1564ae39466ebe982dc2760fa20c1bec0a8

src/lib/y2security/selinux.rb
jreidinger: yep, as default is not there, it is empty string, so fine for me.

src/lib/y2security/selinux.rb
jreidinger: I would be explicit with that nil and also document in in yardoc

dgdavid: Ok, will change.

jreidinger (approved):

src/lib/y2security/selinux.rb
```diff
@@ -145,43 +221,56 @@ def configurable?
...
+
+    # Returns a CFA::Selinux object for handling the config file
     #
-    # @param id [SelinuxConfig::Mode, String, Symbol, nil] a SELinux mode or its identifier
-    # @return [Mode] the SelinuxConfig::Mode by given id or disabled is none found or nil was given
-    def find_mode(id)
-      found_mode = Mode.find(id)
+    # @return [CFA::Selinux]
+    def config_file
+      @config_file ||= CFA::Selinux.load
```
jreidinger, line 243: sadly still exception is not handled. So during installation it cause this crash during save:

selinux_failure

src/lib/y2security/selinux.rb
```diff
@@ -122,21 +188,31 @@ def save
...
 
       Yast::Bootloader.Write
     end
 
+    # Returns needed patterns defined in the product features
+    #
+    # @return [Array<Sring>] collection of defined patterns in product features to have
+    #                        SELinux working as expected
+    def needed_patterns
+      product_feature_settings[:patterns].to_s.split
```
jreidinger, line 204: ah, what about that disabled mode? I am not adapting firewall code and it basically always select it even when selinux is disabled.

src/lib/y2security/selinux.rb
dgdavid: Adressed in https://github.com/yast/yast-security/pull/89, thanks for catching.

src/lib/y2security/selinux.rb
dgdavid: Addressed in https://github.com/yast/yast-security/pull/89, thanks for catching.

PR #88, opened 2021-02-11
AutoYaST for SELinux

dgdavid:
Related to https://github.com/yast/yast-security/pull/83, this PR adds support for SELinux in AutoYaST.

---

<sub>Replaces #85 after _updating_ it to make use of recent changes introduced in yast/yast-security#87</sub>

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -567,7 +567,7 @@ module Yast
     #
     # @return true on success
     def write_selinux
-      return if @Settings["SELINUX_MODE"].empty?
+      return if @Settings["SELINUX_MODE"].to_s.empty?
 
       selinux_config.mode = @Settings["SELINUX_MODE"]
 
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -449,7 +449,8 @@ module Yast
     end
 
     def read_selinux_settings
-      current_mode = Y2Security::Selinux.new.running_mode.id.to_s
+      current_mode = Y2Security::Selinux.new.mode.id.to_s
+
       if current_mode != "disabled"
         @Settings["SELINUX_MODE"] = current_mode
       else
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -84,10 +84,12 @@ module Yast
       Yast.import "UI"
       Yast.import "FileUtils"
       Yast.import "Package"
+      Yast.import "Pkg"
       Yast.import "Pam"
       Yast.import "Progress"
       Yast.import "Service"
       Yast.import "Directory"
+      Yast.import "Report"
       Yast.include self, "security/levels.rb"
     end
 
@@ -446,6 +448,19 @@ module Yast
         "#{@Settings['HIBERNATE_SYSTEM']}"
     end
 
+    def read_selinux_settings
+      selinux = Y2Security::SelinuxConfig.new
+      current_mode = selinux.running_mode.id.to_s
+      if current_mode != "disabled"
+        @Settings["SELINUX_MODE"] = current_mode
+      else
+        @Settings.delete("SELINUX_MODE") if @Settings.has_key?("SELINUX_MODE")
+      end
+
+      log.debug "SELINUX_MODE (after #{__callee__}): " \
+        "#{@Settings['SELINUX_MODE']}"
+    end
+
     # Read all security settings
     # @return true on success
     def Read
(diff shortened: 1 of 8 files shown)


src/modules/Security.rb
```diff
@@ -444,6 +448,16 @@ def read_polkit_settings
         "#{@Settings['HIBERNATE_SYSTEM']}"
     end
 
+    def read_selinux_settings
+      @Settings.delete("SELINUX_MODE")
```
dgdavid, line 452: @schubi2, I didn't catch the idea of using the `@Settings["SELINUX_MODE"]` but I was not brave enough for getting rid of it after adding the convenience method `#selinux_config`, which keep a single instance of `Y2Security::Selinux`. So, if this settings is there just to hold the SELinux mode, I think it can be removed, which would allow to simplify the code.

src/modules/Security.rb
```diff
@@ -547,6 +563,25 @@ def write_shadow_config
...
+      return if @Settings["SELINUX_MODE"].empty?
+
+      selinux_config.mode = @Settings["SELINUX_MODE"]
+
+      # Y2Security::Selinux.mode= fallbacks to the "Disabled" mode when wrong value is given.
+      # In consequence, let's check that the mode still being the same before saving it.
+      if selinux_config.mode.id.to_s != @Settings["SELINUX_MODE"]
+        log.info("The set SELinux mode does not match with the requested one. Not saving it.")
+        false
+      else
```
dgdavid, line 579: As stated there, the `Y2Security::Selinux.mode = :wrong_mode` [will fallback to the disabled mode](https://github.com/yast/yast-security/blob/7c23e55f7af44eae59691c91ef5a65c70c0b497e/src/lib/y2security/selinux.rb#L164-L167). Which most probably is wrong (at least I see it wrong now) and it should be changed to simply do not touch the mode. 

@jreidinger, friendly ping :point_up:

Anyway, probably this extra check (added by me while updating #85) makes no sense at all and we must trust in whatever the _combination_ of  selinux_config.mode= + selinux_config.save does.

jreidinger: well, in this situation when mode is unknown I would use common AutoinstIssues so it report wrong value and list of possible ones. As it is profile error and use later disabled in that case.

dgdavid: > well, in this situation when mode is unknown I would use common AutoinstIssues

If I'm not wrong, this is already done at https://github.com/yast/yast-security/pull/88/commits/222a7a530115870a25c7554ed0069bb22d044b99

So, could we remove this ugly extra check from here safely?

src/modules/Security.rb
dgdavid: Nevermind :) I got it after a few manual tests and also reading https://yastgithubio.readthedocs.io/en/latest/autoyast-development/#autoyast-control-files-old-style 

It's needed to export the value, for example when cloning the system. So, I have only to update the #write_selinux method according to https://github.com/yast/yast-security/pull/88#discussion_r575179956 and maybe exporting it also when is disabled instead of omitting it.

Let's think a little bit more about it.

dgdavid:
@jreidinger, please have a look when you have time.

I tested it manually and I'd say that it works (the system is installed with the suggested SELinux mode). However, I'd prefer you test it too if you don't mind.

The test I did was installing an SLE-15-SP2 using 

* a [minimal AutoYaST profile](https://autoyast.github.io/mini/sles15.xml) including the _selinux\_mode_

  ```xml
  <security>
    <selinux_mode>permissive</selinux_mode>
  </security>
  ```
* and a DUD with the yast-security 4.2.18 RPM and a modified version of skelcd-control-SLES including the _selinux_ configuration

  ```xml
  <globals>
    <selinux>
      <mode>enforcing</mode>
      <configurable>true</configurable>
    </selinux>
  </globals>
  ```

Maybe you can include the new yast-firewall 4.2.6 too.

Thanks in adavance.

dgdavid:
> @jreidinger, please have a look when you have time.

And feel free to send changes if needed. Thanks!

jreidinger:
@dgdavid sadly I worry today I won't have time and tomorrow I won't work, so I can test it on wednesday. If it is too late feel free to merge it and we can handle problems as bug fixes as I expect that this gets some testing by QA.

schubi2:
I would also merge it. This part  does not really depend on the other stuff. 
I have also tested it ;-)

schubi2 (approved):
LGTM

dgdavid:
> I have also tested it ;-)

Thanks a lot :)

PR #90, opened 2021-02-15
Fix popup

jreidinger:
This pull request in yast-yast2 (https://github.com/yast/yast-yast2/pull/1122) changed the behavior during unit tests of `Yast::Report` and `Yast2::Popup`, making necessary to add some extra mocking to prevent YaST from trying to open windows during the test execution.

As far as we know, that affected the test-suites of: yast-storage-ng, yast-country, yast-firewall, yast-nfs-server, yast-nis-client, yast-pam, yast-security, yast-services-manager and yast-sysconfig.

This should fix the problem for this repository.

### DIFF
--- a/test/security_test.rb
+++ b/test/security_test.rb
@@ -241,6 +241,8 @@ module Yast
 
     describe "#write_kernel_settings" do
       before do
+        # there are some failure with running uname in chroot without root perms
+        allow(Yast::Report).to receive(:Error)
         change_scr_root(File.join(DATA_PATH, "system"))
         Security.read_kernel_settings
         stub_scr_write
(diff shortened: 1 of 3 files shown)


test/security_test.rb
```diff
@@ -241,6 +241,8 @@ def enabled?
 
     describe "#write_kernel_settings" do
       before do
+        # there are some failure with running uname in chroot without root perms
+        allow(Yast::Report).to receive(:Error)
```
lslezak, line 245: I think we should rather skipp calling `uname` in the tests than mock the error popup. This might hide some other issues in the code or test. Moreover we should avoid executing some real commands in the system.

jreidinger: it is not so easy as it is hidden very deep in sysctl class - https://github.com/yast/yast-yast2/blob/1629c3a520ebd8e8b93f1161f06d9e140d65b9ec/library/general/src/lib/cfa/sysctl_config.rb#L183
and it handles well if it failed

lslezak (approved):
OK, let's keep it simple...

jreidinger:
thanks

PR #91, opened 2021-02-16
[master] Add SELinux support

dgdavid:
In summary, it adds the SELinux support (also in AutoYaST).

Part of https://trello.com/c/g94xQBDM/ (internal link) and related to work mainly done in https://github.com/yast/yast-security/pull/83, https://github.com/yast/yast-security/pull/87, and https://github.com/yast/yast-security/pull/88.

---

Changes picked from SLE-15-SP2 branch following the [YaST Maintenance work flow](https://yastgithubio.readthedocs.io/en/latest/maintenance-branches/#maintenance-work-flow).

### DIFF
--- a/src/lib/cfa/selinux.rb
+++ b/src/lib/cfa/selinux.rb
@@ -78,8 +78,6 @@ module CFA
       false
     end
 
-    private
-
     # Default path to the SELinux config file
     PATH = "/etc/selinux/config".freeze
     private_constant :PATH
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -449,7 +449,6 @@ module Yast
         "#{@Settings['HIBERNATE_SYSTEM']}"
     end
 
-
     # The name of the PAM module to deal with password quality. Either
     # "pwquality" or "cracklib". See bug #1171318 why this is needed.
     def pwquality_module
@@ -552,7 +551,6 @@ module Yast
       selinux_config.mode = @Settings["SELINUX_MODE"]
       selinux_config.save
     end
-
 
     # Write settings related to PAM behavior
     def write_pam_settings
@@ -786,6 +784,7 @@ module Yast
       end
 
       return true if settings == {}
+
       @modified = true
       tmpSettings = {}
       @Settings.each do |k, v|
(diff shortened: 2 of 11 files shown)


src/lib/cfa/selinux.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    # Default path to the SELinux config file
+    PATH = "/etc/selinux/config".freeze
+    private_constant :PATH
+
+    # The lens to be used by Augeas parser
+    #
+    # @note uses the simplevars lens instead of semanage because the latest is only available from
+    #   augeas-lenses >= 1.12. See https://github.com/hercules-team/augeas/pull/594/files
+    LENS = "simplevars.lns".freeze
```
jreidinger, line 89: is it still valid for master aka SP3?

src/modules/Security.rb
```diff
@@ -772,8 +805,9 @@ def Import(settings)
           end
         end
       end
-      @Settings = tmpSettings
 
+      @Settings = tmpSettings
+      set_selinux_patterns # Checking needed packages
```
jreidinger, line 810: this looks strange. I think autoyast has own method for required packages and not use this PackagesProposal...which I think is not used if confirm mode is set to false.

dgdavid: I didn't know :(

Anyway, since this is "just" a sync PR, lets merge this as it is and fix it first in SP2 and then move it forward to SP3, Do you agree?

src/lib/cfa/selinux.rb
dgdavid: If I'm not wrong, `semanage.lns" is available in SP3 (I have to  recheck it)

src/modules/Security.rb
jreidinger: fine for me.

jreidinger (approved):

src/lib/cfa/selinux.rb
dgdavid: I was wrong :)

The `augeas-lenses` version in SP3 is `1.10.1`

```
$ zypper info augeas-lenses

Information for package augeas-lenses:
--------------------------------------
Repository     : SLE-Module-Basesystem15-SP3-Pool
Name           : augeas-lenses
Version        : 1.10.1-1.11
...
```

PR #96, opened 2021-02-28
Change the unique id for SELinux resolvable

dgdavid:
Change the `SELinux` resolvable unique id used in auto-installation to be consistent with the one used by normal installation.

See https://github.com/yast/yast-firewall/blob/4c1d0ef2db1f982959124186dbd374d722369773/src/lib/y2firewall/clients/proposal.rb#L229-L230

---

Related to https://github.com/yast/yast-security/pull/88 and https://github.com/yast/yast-firewall/pull/144

### DIFF
--- a/src/modules/Security.rb
+++ b/src/modules/Security.rb
@@ -903,7 +903,7 @@ module Yast
     def set_selinux_patterns
       selinux_config.mode = @Settings["SELINUX_MODE"] unless @Settings["SELINUX_MODE"].to_s.empty?
 
-      PackagesProposal.SetResolvables("selinux_patterns", :pattern, selinux_config.needed_patterns)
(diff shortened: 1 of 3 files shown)


jreidinger (approved):

src/modules/Security.rb
```diff
@@ -903,7 +903,7 @@ def default_encrypt_method
     def set_selinux_patterns
       selinux_config.mode = @Settings["SELINUX_MODE"] unless @Settings["SELINUX_MODE"].to_s.empty?
 
-      PackagesProposal.SetResolvables("selinux_patterns", :pattern, selinux_config.needed_patterns)
+      PackagesProposal.SetResolvables("SELinux", :pattern, selinux_config.needed_patterns)
```
jreidinger, line 906: maybe add note and link to the other one so we keep it in sync in future?

jreidinger (approved):

PR #98, opened 2021-03-02
Fix update

jreidinger:
related to https://github.com/yast/yast-installation/pull/923

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Mar  2 11:42:05 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
+- Do not write bootloader in insts-sys (bsc#1182894)
+- 4.3.12
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


dgdavid (approved):
LGTM, thanks for the fix

dgdavid (changes requested):
But, wait. Should we fix it in SP2 first?

jreidinger:
@dgdavid in SLE15 SP2 it is driven by firewall finish which is not called in upgrade - https://github.com/yast/yast-firewall/blob/SLE-15-SP2/src/lib/y2firewall/clients/installation_finish.rb#L54

dgdavid:
> @dgdavid in SLE15 SP2 it is driven by firewall finish which is not called in upgrade - https://github.com/yast/yast-firewall/blob/SLE-15-SP2/src/lib/y2firewall/clients/installation_finish.rb#L54

Ok, but look like those changes could be applied safely in SP2 too, or am I wrong?

dgdavid:
Closed in favor of #100

PR #104, opened 2021-03-03
[SLE-15-SP2] Make SELinux not configurable when running on WSL

dgdavid:
Fix https://bugzilla.suse.com/show_bug.cgi?id=1182940 by making SELinux not configurable when running on [**W**indows **S**ubsystem for **L**inux](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux). Thus, not calling to `Yast::Bootloader` for reading/writing boot kernel params.

> Since SELinux has to be configured via boot kernel params, it makes no
> sense to offer such feature when running on Windows Subsystem for
> Linux[1] becuase the system will be **booted via Windows, so Linux
> bootloader is useless here**[2].
> 
> [1] https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux
> [2] https://yastgithubio.readthedocs.io/en/latest/WSL/

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Mar  3 16:09:26 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Make SELinux not configurable when running on WSL (bsc#1182940)
+- 4.2.23
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


jreidinger:
@dgdavid I think description is right, but as wsl does not have linux kernel, whole selinux does not make sense. So it is fine for me to simply not have it configurable at all.

PR #105, opened 2021-03-03
Make SELinux not configurable when running on WSL

dgdavid:
Fix https://bugzilla.suse.com/show_bug.cgi?id=1182940 by making SELinux not configurable when running on [**W**indows **S**ubsystem for **L**inux](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) since it makes not sense considering that [WSL does not have/use a Linux kernel](https://github.com/yast/yast-security/pull/104#issuecomment-789896900).

---

The same than #104, but for `master` branch.

Trello card (internal link): https://trello.com/c/g5u0hw0j


ancorgs (approved):
LGTM

PR #106, opened 2021-03-18
[SLE-15-SP2] Do not set SELinux mode when it is not configurable

dgdavid:
### Problem

YaST Firstboot is crashing when running in Windows Subsystem for Linux (WSL) because a `Y2Security::Selinux#mode` calls that triggers a `Bootloader#Read`

See https://bugzilla.suse.com/show_bug.cgi?id=1182940

### Solution

Do not `#read_selinux_settings` when SElinux [is not configurable](https://github.com/yast/yast-security/blob/9d96d6e7a8ecc27caa910cc4e5ea7260926e4b3d/src/lib/y2security/selinux.rb#L223-L234).

### Tests

* Added unit tests
* Tested manually
  * Download the Windows 10 image and the appx from openQA.
  * [Enable WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10#manual-installation-steps)
  * [Import the certificate](https://social.msdn.microsoft.com/Forums/en-US/b9281639-5bb4-4064-a5dc-c769183d7d91/new-certificate-must-be-installed-for-this-app-package)
  * Install the appx.
  * Launch the appx takes you to a console. Start yast2-firstboot then: `/usr/lib/YaST2/startup/YaST2.call firstboot firstboot`.

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Mar 18 11:43:42 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Do not set SELinux mode when it is not configurable (bsc#1182940)
+- 4.2.24
+
+-------------------------------------------------------------------
 Wed Mar  3 16:09:26 UTC 2021 - David Diaz <dgonzalez@suse.com>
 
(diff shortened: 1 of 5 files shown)


ancorgs (approved):
LGTM

dgdavid:
IBS request: https://build.suse.de/request/show/238192

PR #109, opened 2021-04-13
Merge SLE-15-SP2 into SLE-15-SP3

imobachgs:
Merge #108 into `master`. No real changes, just preventing introducing a bug when merging in the future.


lslezak (approved):
:+1:

PR #111, opened 2021-07-14
Read without exceptions

joseivanlopez:
## Problem

The `Security` module offers a `Read` method which raises an exception if reading the settings fails (e.g., parsing error, file does not exist, etc). But that exceptions are lost when the `Security#Read` is called from a perl module like `Users.pm`. As result, the perl code has no way to check if there was an error while reading the security settings.

* Related to https://bugzilla.suse.com/show_bug.cgi?id=1177183
* Needed for https://github.com/yast/yast-users/pull/317

## Solution 

Add a new method `Security#SafeRead` which avoids to raise exceptions. It stores the error message when reading the settings fails.

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jul 14 13:51:18 UTC 2021 - José Iván López González <jlopez@suse.com>
+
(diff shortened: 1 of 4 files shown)


imobachgs (approved):

joseivanlopez:
https://build.suse.de/request/show/245204

PR #115, opened 2021-12-15
Add support for selecting desired Linux Security Module (LSM)

teclator:
## Problem

It has been requested to unify the [Linux Security Module (**LSM**)](https://www.kernel.org/doc/html/latest/admin-guide/LSM/index.html) selection during installation, allowing not only to select [**SELinux**](https://selinuxproject.org/page/Main_Page) but also [**AppArmor**](https://apparmor.net/) or even none **LSM** major module at all.

- https://trello.com/c/0KpnKW6N/2754-8-feature-shouldhave-unify-selinux-and-apparmor-solution-in-the-installation-summary

## Solution

In order to choose which **LSM** should be used by default we have added a new section `lsm` to the control file as well as we have added almost the same options to the security **AutoYaST** schema. The default LSM will be set by the `select` tag falling back to **apparmor** in case of missing.

```xml

<lsm>
  <select>selinux</select>
  <configurable config:type="boolean">true</configurable>

  <apparmor>
    <patterns>apparmor</patterns>
    <selectable config:type="boolean">true</selectable>
  </apparmor>
  
  <selinux>
    <!-- Set SELinux enforcing mode by default -->
    <mode>enforcing</mode>
    <configurable config:type="boolean">true</configurable>
    <selectable config:type="boolean">true</selectable>
    <patterns>selinux</patterns>
  </selinux>
</lsm>

```

The **configurable** option is available in the `lsm` section but also per module. If the option is defined as false in the **lsm** section then the **LSM** configuration or dialog will not be exposed.

The **selectable** options is only available per module and will omit the module from the select list during the installation.

## See also

* https://github.com/yast/yast-installation/pull/1006

## Screenshots

| Dialog | Screenshot |
| --------- | --------------- |
| **Proposal - Selinux Permissive** | YastSecurity_SELinuxPermissive |
| **Proposal - AppArmor** | YastSecurity_AppArmor |
| **Proposal - No LSM major module wanted** | YastSecurity_None |
| **Security Proposal Dialog - LSM Selector** | YastSecurity_LSMSelector |
| **Security Proposal Dialog - Selector with Selinux in Permissive mode** | YastSecurity_SelinuxPermissiveSelector |

PR #117, opened 2022-01-04
Do not propose any Linux Security Module if declared as not configurable

teclator:
In case that the **Linux Security Module** should not be configured during the installation then we should not select any module at all in order to not write the boot parameters.

Related to #115

After discussing the current behavior with @dgdavid it probably makes sense to have also a differentiation between **selectable** and **configurable** for the **LSM** config, adding the **selectable** tag also there.

In case that LSM is not selectable then the widget and probably the proposal could be hidden but we could select a default LSM in case of configurable. If it is marked as no configurable then it will be hidden and also skipped completely.


dgdavid (approved):
LGTM

PR #118, opened 2022-01-07
Use security

teclator:
## Problem

As it seems that we need to specify also the **integrity** module in order to be able to use the **Linux Integrity Measurement Architecture (IMA)** subsystem it has been decided that by now it is preferable to continue using the **security** boot parameter instead of the **lsm** one.

See https://github.com/openSUSE/kernel-source/commit/953db3507d62938038cd5f2d4bbe7fd0cb3d3763

- https://bugzilla.suse.com/show_bug.cgi?id=1194332

**Note**: We could write **lsm=integrity,apparmor** when **apparmor** is selected having the same value we have by default but that looks strange and we should add it in case we allow to select a list of modules to be set during installation.

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Jan  7 12:37:08 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
+- Do not use the 'lsm' kernel boot parameter by now as it could
+  need other modules to be pecified like the integrity one
+  (bsc#1194332).
+- 4.4.5
+
+-------------------------------------------------------------------
 Tue Jan  4 12:06:48 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
 
 - Related to jsc#SLE-22069:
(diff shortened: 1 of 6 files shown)


dgdavid (approved):
LGTM.

For completeness, just remind that the decision of going for `lsm` instead of `security` boot kernel param was taken based on the information written in the [kernel admin guide](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html):

> security=       [SECURITY] Choose a legacy "major" security module to
                        enable at boot. **This has been deprecated by the
                        "lsm=" parameter**.

However, according to the same documentation, `lsm` is overriding not only `security` but also `CONFIG_LSM` 

>  lsm=lsm1,...,lsmN
                       [SECURITY] Choose order of LSM initialization. **This
                       overrides CONFIG_LSM**, and the "security=" parameter.

and the _integrity_ module is currently initialized becauset its inclusion in [CONFIG_LSM](https://github.com/openSUSE/kernel-source/commit/953db3507d62938038cd5f2d4bbe7fd0cb3d3763)

```diff
-CONFIG_LSM=""
+CONFIG_LSM="integrity,apparmor"
```

package/yast2-security.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Fri Jan  7 12:37:08 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
+- Do not use the 'lsm' kernel boot parameter by now as it could
+  need other modules to be pecified like the integrity one
```
pevik, line 5: typo: s/pecified/specified/

dgdavid: Thanks @pevik 

I caught the typo this morning too, but the change was sent a month ago or so.

PR #119, opened 2022-01-10
Fix security auto requirement

teclator:
## Problem

The requirement was not updated since we moved the **SELinux** class under **LSM** (related to #115)

- https://bugzilla.suse.com/show_bug.cgi?id=1194449

## Solution

Fix the dependency

## Tests

Tested manually but we probably should add a unit test for the auto client **ASAP**.


dgdavid (approved):
LGTM. Thanks for your quick patch!

PR #120, opened 2022-01-11
Autoyast LSM section: Added 'none' subsection allowing to mark the module as not selectable

teclator:
It will allow to hide the option in case of not wanted:

```xml
<?xml version="1.0"?>
<!DOCTYPE profile>
<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
  <security>
  <lsm>
    <select>apparmor</select>
    <configurable config:type="boolean">true</configurable>
    <none>
      <!-- The option will not be available in the confirmation proposal for selecting it -->
      <selectable config:type="boolean">false</selectable>
    </none>
  </lsm>
  </security>	
</profile>
```

related to #115

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Jan 11 00:06:20 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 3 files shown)


dgdavid (approved):
LGTM

PR #123, opened 2022-01-26
AutoYaST LSM: limited behavior

teclator:
## Problem

During the review of https://trello.com/c/0KpnKW6N/2754-8-feature-shouldhave-unify-selinux-and-apparmor-solution-in-the-installation-summary it was suggested to not expose the same options available in the installation control file but only allow the user to select the desired **LSM** as well as the **SELinux** mode in case of selected.


src/lib/y2security/lsm/config.rb
```diff
@@ -145,6 +145,16 @@ def configurable?
         @configurable = product_feature_settings.fetch(:configurable, true)
       end
 
+      # Export AutoYaST LSM configuration
+      #
+      # @return [Hash<String, String]
+      def export
+        config = {}
+        config["lsm_select"] = selected&.id.to_s if selected
```
dgdavid, line 153: NP: You're already checking `if selected`. So, `&.` is not needed here.

test/y2security/autoinst_profile/security_section_test.rb
```diff
@@ -18,36 +18,17 @@
...
-      expect(lsm.parent).to eq(section)
-    end
-
-    context "when used the old 'selinux_mode' attribute" do
-      let(:profile) { { "selinux_mode" => "enforcing" } }
+    context "when only 'selinux_mode' attribute is given" do
+      let(:profile) { { "selinux_mode" => "enforcing", "lsm_select" => "selinux" } }
 
-      it "sets the selinux_mode attribute" do
+      it "sets the supported attributes" do
```
dgdavid, line 28: NP: I'd say "it selects SELinux as LSM"

test/y2security/lsm/config_test.rb
```diff
@@ -258,4 +258,29 @@
...
+    end
+
+    context "when a LSM is selected" do
+      it "exports the selected LSM" do
+        subject.select("apparmor")
+        expect(subject.export).to eql("lsm_select" => "apparmor")
+      end
+
+      context "and it is SELinux" do
+        it "exports the SELInux mode" do
```
dgdavid, line 276: NP: SEL**i**nux

dgdavid (approved):
Despite those minor comments, it LGTM :+1:

package/yast2-security.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-security
-Version:        4.4.9
+Version:        4.4.10
```
jreidinger, line 20: do not forget to bump also schema spec version as you modify schema, so it has to be regenerated

src/lib/y2security/autoinst/lsm_config_reader.rb
```diff
@@ -18,55 +18,47 @@
...
     private
 
-      def configure_supported_modules
-        [:selinux, :apparmor, :none].each do |id|
-          lsm_module = config.public_send(id)
-          @module_section = section.public_send(id)
-          next unless module_section
+      def selinux?
+        return true if section.lsm_select == "selinux"
+        return true if !section.lsm_select && section.selinux_mode
```
jreidinger, line 50: I expect rubocop will complain that you can write it simply as `!section.lsm_select && section.selinux_mode` without return, if and false

src/autoyast-rnc/security.rnc
```diff
@@ -9,7 +9,9 @@ cwd_in_user_path = element cwd_in_user_path { STRING }
 disable_restart_on_update = element disable_restart_on_update { STRING }
 disable_stop_on_removal = element disable_stop_on_removal { STRING }
 extra_services = element extra_services { STRING }
-selinux_mode = element selinux_mode { STRING }
+selinux_mode = element selinux_mode { "permissive" | "enforcing" | "disabled" }
```
jreidinger, line 12: so is it still deprecated at https://github.com/yast/yast-security/blob/master/src/lib/y2security/autoinst_profile/security_section.rb#L44 ?

src/lib/y2security/autoinst/lsm_config_reader.rb
jreidinger: and please explain why ( I expect backward compatibility? or reasonable default? )

src/lib/y2security/autoinst/lsm_config_reader.rb
```diff
@@ -18,55 +18,47 @@
...
-          assign(lsm_module, :selectable)
-          next if id == :none
+        false
+      end
 
-          assign(lsm_module, :configurable)
-          assign(lsm_module, :patterns)
-        end
+      def configure_selinux
+        config.selinux.mode = section.selinux_mode
```
jreidinger, line 56: what if section is missing mode. Is it ok to have it as nil?

src/autoyast-rnc/security.rnc
jreidinger: ah never mind. I overlook that here is also modification of that security section.

src/lib/y2security/autoinst_profile/security_section.rb
```diff
@@ -18,53 +18,31 @@
...
-    #       <selectable config:type="boolean">false</selectable>
-    #     </apparmor>
-    #     <selinux>
-    #       <mode>permissive</mode>
-    #       <configurable config:type="boolean">true</configurable>
-    #       <patterns>selinux</patterns>
-    #     </selinux>
-    #   </lsm>
+    #   <selinux_mode>enforcing</selinux_mode>
+    #   <lsm_select>apparmor</lsm_select>
```
jreidinger, line 29: does example makes sense at all? with apparmor I expect that selinux mode is not selected at all and maybe show warning?

test/y2security/lsm/config_test.rb
jreidinger: I would say "exports also the SELinux mode" to make it obvious that selected lsm is also exported.

package/yast2-security.spec
teclator: Yep, it is here https://github.com/yast/yast-schema/pull/127

src/autoyast-rnc/security.rnc
teclator: yep, basically we introduced the **lsm** section in #115 but after the review it was decided to no expose so much functionality as it is done in the installation control file, so, as we only offer the option for selecting the desired **LSM** we also removed the new section

src/lib/y2security/autoinst/lsm_config_reader.rb
teclator: > I expect rubocop will complain that you can write it simply as `!section.lsm_select && section.selinux_mode` without return, if and false

Rubocop does not complain but you are right.

teclator: > and please explain why ( I expect backward compatibility? or reasonable default? )

I would say that It is the same behavior than before #115 was introduced. If no LSM is selected explicitly then it will select SELinux if the **selinux_mode** is declared.

src/lib/y2security/autoinst/lsm_config_reader.rb
teclator: Yes it is ok. If there is no section then the defaults will be computed during the proposal which will try to select the default **LSM** based on the control file and the same happens with the SELinux mode if it is not set explicitly.

See https://github.com/yast/yast-security/blob/master/src/lib/y2security/lsm/selinux.rb#L121

src/lib/y2security/autoinst_profile/security_section.rb
teclator: No you are right. The example does not make sense and I added for showing both options but should use **SELinux** in this case.

We were discussing about the different options. The user can define both as the example shows, and that would influence the default **SELinux** mode exposed in the proposal but AppArmor will be selected, so, only if the selection is modified the value will be relevant and @dgdavid preferred to just ignore the option if the **LSM** selected explicitly is different to **SELinux**

I guess that we could see it as the **lsm** kernel option which has precedence over **security** parameter and which invalidates any **SELinux** option if it is not the **major** module selected.

test/y2security/autoinst_profile/security_section_test.rb
teclator: I forget to remove the upper context. It does not select nothing just set the attributes as are defined, the select is done by the reader.

jreidinger (approved):

PR #125, opened 2022-02-22
Stop using the 'lsm' kernel boot parameter

dgdavid:
Even for the _None_ option, which was forgotten when going back from the recommended `lsm` to the `security` one. See https://github.com/yast/yast-security/pull/118.

This should fix https://bugzilla.suse.com/show_bug.cgi?id=1196274

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Feb 22 08:08:56 UTC 2022 - David Diaz <dgonzalez@suse.com>
+
+- Stop using 'lsm' kernel boot parameter even for the
+  "None" Major Linux Security Module (bsc#1194332, bsc#1196274).
+- 4.4.12
+
+-------------------------------------------------------------------
 Fri Feb  4 09:19:19 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
 
 - Fixed Export unit test (related to jsc#SLE-22069).
--- a/package/yast2-security.spec
+++ b/package/yast2-security.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-security
-Version:        4.4.11
+Version:        4.4.12
 Release:        0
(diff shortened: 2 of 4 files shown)


lslezak (approved):

teclator:
@dgdavid if it disables **AppArmor** / **SELinux** then it is fine with me

dgdavid:
> @dgdavid if it disables **AppArmor** / **SELinux** then it is fine with me

Good catch!

Honestly, yesterday I didn't test it manually because early in the morning the solution appeared straight forward to me. But it is true that we were speaking about that during the implementation of "None" option. Sadly, I forget about it and at the time to revisit the code I simply though "Dam! We overlooked that one"

But it is true that the kernel documentation says

> If AppArmor is not the default security module it can be enabled by passing security=apparmor on the kernel’s command line.
> 
> **If AppArmor is the default security module** it can be disabled by passing **apparmor=0, security=XXXX (where XXXX is valid security module)**, on the kernel’s command line.

https://www.kernel.org/doc/html/latest/admin-guide/LSM/apparmor.html

And yes, AppArmor is the default security module in our kernel. So, according to that documentation, looks difficult to go for a "None" security module. 

However, I have tested the current fix manually and… it works (see some screenshots at the end of the comment). Although, admittedly, it is not so elegant having a `security=` boot param there. But we can use either, `security=0` or `security=none`. We can agree that they are not perfect, but at least are _more expressive_ than the current one, especially the last one.

What do you think?

---

| Description | Screenshot |
|-|-|
| openSUSE MircroOS using SELinux enforcing mode |  microos-selinux-enforcing |
| SLE-15-SP3 using AppArmor by default (no `security` boot param) | sle-15-sp3-apparmor |
| SLE-15-SP3 using `security=` | empty_security_kernel_param |
| SLE-15-SP3 using `security=0` | sle-15-sp3-security_to_zero | 
| SLE-15-SP3 using `security=none` | sle-15-sp3-security-none |

PR #126, opened 2022-03-11
Force package backend

imobachgs:
Use https://github.com/yast/yast-yast2/pull/1249 to avoid potential problems like the one described in [bsc#1196963](https://bugzilla.suse.com/show_bug.cgi?id=1196963).


ancorgs (approved):
LGTM

PR #128, opened 2022-08-05
Add a security policy validation mechanism

imobachgs:
This PR adds security policy validation to the installer (see https://www.open-scap.org/security-policies/choosing-policy/).

- https://jira.suse.com/browse/PM-3526
- https://trello.com/c/SqFGWTyB
- https://trello.com/c/wcwOKwjK
- https://trello.com/c/4rwfhcTt

Related PRs:
  * https://github.com/yast/yast-storage-ng/pull/1308.
  * https://github.com/yast/yast-autoinstallation/pull/845
  * https://github.com/yast/yast-installation/pull/1059
  * #131
  * #132
  * #133
  * #134
  * #135
  * #136
  * #138
  * https://github.com/yast/skelcd-control-leanos/pull/96
  
## How it works

See the screenshots below to get an idea of how it works. Once the user enables a security profile (at this point only [DISA STIG](http://static.open-scap.org/ssg-guides/ssg-sle15-guide-stig.html)), YaST:

* Add the required package (`ssg-apply`) and enable the service at the end of the installation.
* Runs a set of checks and displays the rules that are not passing. 
  * If the problem can be fixed automatically, it displays a link to do it.
  * Otherwise, it might display a link to take the user to the right place of the installer to solve it.

<details>
  <summary>Installation settings including a 'Security Policy' section</summary>

stig-installation-settings
</details>

<details>
  <summary>Storage proposal showing found problems with the current configuration</summary>

stig-installation-summary
</details>

<details>
  <summary>YaST warning about an issue in the expert partitioner</summary>

stig-partitioner
</details>

<details>
  <summary>AutoYaST confirmation mode when some rule failed</summary>

stig-autoyast-confirm-mode
</details>

<details>
  <summary>Do not allow installing the system until all problems are solved</summary>

stig-enabled-blocked
</details>

## Enabling security policy validation

There are three different ways to enable policy checks:

* Clicking "enable" on the installation summary page.
* Passing the option `YAST_SECURITY_POLICIES=stig` at boot time.
* Using AutoYaST (see https://github.com/yast/yast-autoinstallation/pull/845 for further details).

## What is missing?

Write the name of the enabled policies and disabled rules to the file system, so `ssg-apply` can take that information into account.

## Implementation details

* The API is defined in the [Y2Security::SecurityPolicies](https://github.com/yast/yast-security/tree/security-policy/src/lib/y2security/security_policies) module.
[...]

PR #131, opened 2022-09-06
Use a class per each rule and remove issues/actions

imobachgs:
Based on #128.

PoC to use a class per security policy rule.

Screenshot from 2022-09-13 10-25-30

# See also

* https://github.com/yast/yast-security/pull/132


src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -323,6 +247,164 @@ def storage_client
...
+        def build_link(action, id = nil)
+          "#{dialog_id}--#{action}:#{id}"
+        end
+      end
+
+      # Builds the representation of a security policy
+      class PolicyPresenter
+        include Yast::I18n
+
+        def initialize(policy, enabled:, failing_rules:, links_builder:)
```
mvidner, line 302:
```suggestion
        # @param policy [SecurityPolicies::Policy]
        # @param failing_rules [Array<SecurityPolicies::Rule>]
        # @param links_builder [LinksBuilder]
        def initialize(policy, enabled:, failing_rules:, links_builder:)
```

failing_rules is a bit confusing because elsewhere it is a hash of arrays

src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -109,118 +120,75 @@ def ask_user(param)
...
-      # All policies
+      # Runs the security policies checks
       #
-      # @return [Array<Y2Security::SecurityPolicies::Policy>]
-      def policies
-        policies_manager.policies
+      # Calling this method updates the list of issues in
+      # Y2Security::Clients::SecurityProposalProposal.issues.
+      #
+      # @see Y2Security::SecurityPolicies::Manager#issues
```
mvidner, line 158: the classes referenced above are outdated and we should be mentioning *failing_rules*, right?

src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -233,80 +201,36 @@ def refresh_packages
...
+        Y2Security::SecurityPolicies::Manager.instance
       end
 
-      # Returns the list of issues
-      #
-      # The reason to hold a variable with the list is that issues are identified by its position in
-      # the list, so we need to be sure that the list does not change.
-      def all_issues
-        @all_issues ||= issues.all
+      def target_config
```
mvidner, line 232:
```suggestion
      # @return [SecurityPolicies::TargetConfig]
      def target_config
```

src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -323,6 +247,164 @@ def storage_client
...
+        def initialize(policy, enabled:, failing_rules:, links_builder:)
+          textdomain "security"
+
+          @policy = policy
+          @policy_enabled = enabled
+          @failing_rules = failing_rules
+          @links_builder = links_builder
+        end
+
+        def to_html
```
mvidner, line 311:
```suggestion
        # @return [String]
        def to_html
```

src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -323,6 +247,164 @@ def storage_client
...
+
+          Yast::HTML.List(items)
+        end
+      end
+
+      # Builds the representation of a rule
+      class RulePresenter
+        include Yast::I18n
+
+        def initialize(rule, links_builder)
```
mvidner, line 358:
```suggestion
        #  @param rule [SecurityPolicies::Rule]
        def initialize(rule, links_builder)
```

src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -323,6 +247,164 @@ def storage_client
...
+
+        def actions
+          rule_actions = [toggle_action]
+          rule_actions << fix_action if rule.enabled? && rule.fixable?
+
+          rule_actions
+        end
+
+        def toggle_action
+          text = rule.enabled? ? _("disable rule") : _("enable rule")
```
mvidner, line 390: use an `if` to put them on separate lines and add TRANSLATOR comments

mvidner (commented):
I read the proposal part and added the types as I figured them out.

src/lib/y2security/clients/security_policy_proposal.rb
mvidner: Now that I think of it, single line may work just fine

src/lib/y2security/security_policies/manager.rb
```diff
@@ -88,20 +91,12 @@ def enabled_policy?(policy)
...
-      # The issues can be limited to a specific scope.
-      #
-      # @param scope [Scopes::Storage, Scopes::Bootloader, Scopes::Network, Scopes::Firewall]
-      # @return [IssuesCollection]
-      def issues(scope = nil)
-        issues_collection = IssuesCollection.new
-
-        enabled_policies.each do |policy|
-          issues_collection.update(policy, policy.validate(scope))
+      # @return [Hash<Policy, Array<Rule>>]
```
mvidner, line 94:
```suggestion
      # @param config [TargetConfig]
      # @param scope [Symbol,nil] only consider rules with this scope.
      #   For example yast2-storage-ng will call this with :storage to
      #   only display rules that it can fix.
      # @return [Hash<Policy, Array<Rule>>]
```

We got rid of scope classes but I wondered why still have the argument. Now I know.

mvidner (changes requested):

src/lib/y2security/security_policies/missing_encryption_rule.rb
```diff
@@ -0,0 +1,59 @@
...
+require "y2storage"
+
+module Y2Security
+  module SecurityPolicies
+    # Rule to check that all file systems are encrypted (except /boot/efi) (SLES-15-010330).
+    class MissingEncryptionRule < Rule
+      def initialize
+        textdomain "security"
+
+        super("SLES-15-010330", _("Separate file systems must be encrypted"), :storage)
```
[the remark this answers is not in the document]
imobachgs: I have updated the description using your second proposal.

mvidner (approved):

PR #134, opened 2022-09-14
Enable/disable the ssg-apply service according to the security policies

imobachgs:
- https://trello.com/c/H2WYkfZf/3081-3-stig-trigger-check-on-first-boot-with-persisted-list-of-disabled-rules

* When a security policy is enabled, set the `ssg-apply` package and service for installation.
* When no security policies are enabled, remove the `ssg-apply` package and service.

### DIFF
--- a/src/lib/y2security/security_policies/manager.rb
+++ b/src/lib/y2security/security_policies/manager.rb
@@ -167,6 +167,10 @@ module Y2Security
       end
 
       # Return the list of enabled services
+      #
+      # FIXME: avoid a cyclic dependency with yast2-installation
+      #
+      # @return [Array<String>] List of enabled services
       def enabled_services
         require "installation/services" unless defined?(::Installation::Services)
         ::Installation::Services.enabled
--- a/src/lib/y2security/security_policies/manager.rb
+++ b/src/lib/y2security/security_policies/manager.rb
@@ -21,6 +21,8 @@ require "yast"
 require "singleton"
 require "y2security/security_policies/disa_stig_policy"
 
+Yast.import "PackagesProposal"
+
 module Y2Security
   module SecurityPolicies
     # Class to manage security policies
@@ -74,13 +76,17 @@ module Y2Security
         return unless policies.include?(policy)
 
         @enabled_policies.push(policy).uniq!
+        enable_service
       end
 
       # Disables the given policy
       #
       # @param policy [Policy]
       def disable_policy(policy)
+        return unless policies.include?(policy)
+
         @enabled_policies.delete(policy)
+        disable_service if @enabled_policies.empty?
       end
 
       # Whether the given policy is enabled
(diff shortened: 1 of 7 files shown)


src/lib/y2security/security_policies/manager.rb
```diff
@@ -140,6 +146,35 @@ def env_policies
...
+
+      SERVICE_NAME = "ssg-apply".freeze
+      private_constant :SERVICE_NAME
+
+      # Adds the package and enables the service to remedy the system after the installation
+      def enable_service
+        return if enabled_services.include?(SERVICE_NAME)
+
+        enabled_services << SERVICE_NAME
+        Yast::PackagesProposal.AddResolvables("security", :package, [SERVICE_NAME])
```
mvidner, line 158: Where is the package coming from? Last I checked, it was in a special add-on product

imobachgs: At this point, I do not know. I would expect the software solver to fail if the package is not available (I added the package in a DUD).

src/lib/y2security/security_policies/manager.rb
```diff
@@ -140,6 +146,35 @@ def env_policies
...
+      def disable_service
+        return unless enabled_services.include?(SERVICE_NAME)
+
+        enabled_services.delete(SERVICE_NAME)
+        Yast::PackagesProposal.RemoveResolvables("security", :package, [SERVICE_NAME])
+      end
+
+      # Return the list of enabled services
+      def enabled_services
+        require "installation/services" unless defined?(::Installation::Services)
```
joseivanlopez, line 171: I would add a comment about cyclic dependency problem, similar to https://github.com/yast/yast-security/pull/128/files#diff-b6b853b94bce3cd38105d93a59732aa0345ed6ad7e53df5afad8e337d13bdb7cR81.

joseivanlopez: BTW, yast2-installation only has a BuildRequires yast2-security. Could yast2-security require yast2-installation?

joseivanlopez: And I would say that `SecuritySettings` and `Services` should not be in yast2-installation.

test/y2security/security_policies/manager_test.rb
```diff
@@ -91,6 +91,11 @@
   end
 
   describe "#enable_policy" do
+    before do
+      allow(subject).to receive(:require).with("installation/services")
```
joseivanlopez, line 95: I don't get this. The require is mocked but the class `Installation::Services` is used below. Does it actually work if the require is mocked?

imobachgs: Yes, because we are kind of mocking that class (see https://github.com/yast/yast-security/pull/134/files#diff-ba37813ca277c227a74a372479b7b05b7f3ff085d890ab708f80d62573efdb7a).

src/lib/y2security/security_policies/manager.rb
imobachgs: I agree. I guess they were placed there for some reason. Perhaps we might consider changing them for SP5, but for SP4 I would keep them as they are.

joseivanlopez (approved):
LGTM

PR #135, opened 2022-09-14
Improve DISA STIG rules

joseivanlopez:
## Problem

* [SLES-15-040210](http://static.open-scap.org/ssg-guides/ssg-sle15-guide-stig.html#xccdf_org.ssgproject.content_rule_partition_for_var_log_audit) The system has a separate file system for /var/log/audit ([stigviewer](https://www.stigviewer.com/stig/suse_linux_enterprise_server_15/2022-06-06/finding/V-234980)).

* [SLES-15-030660](http://static.open-scap.org/ssg-guides/ssg-sle15-guide-stig.html#xccdf_org.ssgproject.content_rule_auditd_audispd_configure_sufficiently_large_partition) File system /var/log/audit must have enough capacity for audit records ([stigviewer](https://www.stigviewer.com/stig/suse_linux_enterprise_server_15/2022-06-06/finding/V-234965)).

* [SLES-15-010190](http://static.open-scap.org/ssg-guides/ssg-sle15-guide-stig.html#xccdf_org.ssgproject.content_group_non-uefi) A bootloader password (for grub2) must be configured (BIOS) ([stigviewer](https://www.stigviewer.com/stig/suse_linux_enterprise_server_15/2022-02-11/finding/V-234819)).

## Solution

*Short description of the fix.*

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

*If the fix affects the UI attach some screenshots here.*


src/lib/y2security/security_policies/bootloader_password_rule.rb
```diff
@@ -18,17 +18,22 @@
...
     #  Run to check whether grub2 is password-protected and menu editing is restricted
-    #  (SLES-15-010200).
     class BootloaderPasswordRule < Rule
       def initialize
         textdomain "security"
 
+        # DISA STIG defines a rule for UEFI (SLES-15-010200) and another rule for non-UEFI
+        # (SLES-15-010190). The condition to check both rules in YaST is exactly the same, so let's
+        # simply adapt the rule ID according to the system type.
+        id = Y2Storage::Arch.new.efiboot? ? "SLES-15-010200" : "SLES-15-010190"
```
jreidinger, line 33: btw in bootloader you can redefine type of bootloader to use non-efi on efi and vice versa, but usually it is not good idea ( I know that it was useful for some strange archs where medium cannot be efi booted, but still there is issue with efi manager which cannot access efi menu), so I would ignore it here.

src/lib/y2security/security_policies/filesystem_size_rule.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      # Minimum size the file system should have
+      #
+      # @return [Y2Storage::DiskSize]
+      attr_reader :min_size
+
+      # @param id [String] Rule ID
+      # @param mount_path [String] Mount path for the file system to check
+      # @param min_size [Y2Storage::DiskSize] Minimum size the file system should have
+      def initialize(id, mount_path, min_size)
```
jreidinger, line 39: I would use here for `min_size` keyword parameter as in future it can easily exist rule about max size or another condition on size like size range. So I would not force to always use min_size...and also calling it will look better. Compare `FilesystemSizeRule.new("id5", "/mnt", Y2Storage::DiskSize.MiB(100))` versus `FilesystemSizeRule.new("id5", "/mnt", min_size: Y2Storage::DiskSize.MiB(100))`. So as long as rule is SizeRule and not MinSizeRule I would use keyword parameter.

jreidinger (approved):

PR #136, opened 2022-09-15
Block installation if security policy checks fail

imobachgs:
Block the installation when a security policy check fails.

stig-installation-summary

### DIFF
--- a/src/lib/y2security/clients/security_policy_proposal.rb
+++ b/src/lib/y2security/clients/security_policy_proposal.rb
@@ -73,7 +73,7 @@ module Y2Security
           "preformatted_proposal" => preformatted_proposal,
           "links"                 => links,
           "warning_level"         => warning_level,
-          "warning"               => nil
+          "warning"               => warning
         }
       end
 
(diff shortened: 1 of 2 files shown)


src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -128,17 +128,30 @@ def links
         links.flatten.compact.uniq
       end
 
+      # Returns the warning message
+      #
+      # @return [String,nil] warning message or nil if there are no failing rules
+      def warning
+        return nil if success?
+
+        _("The current configuration does not comply with the enabled security policies")
```
mvidner, line 137: The language is fine.

What about a period `.` at the end? Do other messages in this context include it?

imobachgs: You are right: e.g., we are using a period in the software proposal. I have added it to the message.

src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -128,17 +128,30 @@ def links
...
       # Blocker if there are enabled failing rules
       #
       # @return [Symbol] :blocker
       def warning_level
-        rules = failing_rules.values.flatten
-
-        return :warning if rules.none? || rules.none?(&:enabled?)
-
-        :blocker
+        success? ? :warning : :blocker
```
mvidner, line 154: From [ProposalClient#make_proposal](https://www.rubydoc.info/github/yast/yast-yast2/master/Installation/ProposalClient#make_proposal-instance_method) I understand that `warning_level` is only relevant if `warning` is not `nil`

```suggestion
        success? ? nil : :blocker
```

mvidner (approved):

PR #137, opened 2022-09-19
Write `ssg-apply` configuration file

imobachgs:
The `ssg-apply` tool read its configuration from the `default.conf` and `override.conf` files in the`/etc/ssg-apply` directory. The idea of this PR is to write the configuration in the latter of those files, including the policy to apply and the rules to ignore.

This PR also contains some fixes in the policies proposal.

### DIFF
--- a/src/lib/cfa/ssg_apply.rb
+++ b/src/lib/cfa/ssg_apply.rb
@@ -47,18 +47,58 @@ module CFA
       profile: "profile", disabled_rules: "disabled-rules"
     )
 
+    class << self
+      # Loads a file
+      #
+      # @param file_handler [#read,#write] an object able to read/write a string (like File)
+      # @param file_path    [String] File path
+      # @return [SsgApply] File with the already loaded content
+      def load(file_handler: Yast::TargetFile, file_path: PATH)
+        file = new(file_handler: file_handler, file_path: file_path)
+        file.load
+        file
+      rescue Errno::ENOENT
+        log.info("#{file_path} couldn't be loaded. Probably the file does not exist yet.")
+
+        file
+      end
+    end
+
+    # @param file_handler [#read,#write] an object able to read/write a string (like File)
+    # @param file_path    [String] File path
     def initialize(file_handler: Yast::TargetFile, file_path: PATH)
       super(AugeasParser.new(LENS), file_path, file_handler: file_handler)
     end
 
+    # Returns the list of disabled rules
+    #
+    # To add rules to the list, please use the #disabled_rules= method instead of
+    # adding rules to the value returned by this method.
+    #
+    # @return [Array<String>] List of disabled rules
     def disabled_rules
       @disabled_rules ||= generic_get("disabled-rules")
         .to_s.split(",").freeze
     end
 
+    # Sets the list of disabled rules
+    #
+    # @example Adding disabled rules
+    #   file = SsgApply.load
+    #   rules = file.disabled_rules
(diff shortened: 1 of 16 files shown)


src/lib/cfa/ssg_apply.rb
```diff
@@ -0,0 +1,104 @@
...
+    # To add rules to the list, please use the #disabled_rules= method instead of
+    # adding rules to the value returned by this method.
+    #
+    # @return [Array<String>] List of disabled rules
+    def disabled_rules
+      @disabled_rules ||= generic_get("disabled-rules")
+        .to_s.split(",").freeze
+    end
+
+    # Sets the list of disabled rules
```
mvidner, line 84:
```suggestion
    # Sets the list of disabled rules
    # @param value [Array<String>] List of disabled rules
```

src/lib/cfa/ssg_apply.rb
```diff
@@ -0,0 +1,104 @@
...
+    #
+    # @example Adding disabled rules
+    #   file = SsgApply.load
+    #   rules = file.disabled_rules
+    #   new_rule = Rule.new("SLES-15-000000", "My dummy rule", :unknown)
+    #   file.disabled_rules = rules + [new_rule]
```
mvidner, line 90: Conceptually the disabled rules are a set. Maybe it would be easier to use a `Set` class instead of an `Array`?

src/lib/cfa/ssg_apply.rb
```diff
@@ -0,0 +1,104 @@
...
+    def disabled_rules=(value)
+      @disabled_rules = nil
+      generic_set("disabled-rules", value.join(","))
+    end
+
+    # Determines whether the file is empty
+    #
+    # @return [Boolean] true if it is empty; false otherwise
+    def empty?
+      data.data.empty?
```
mvidner, line 101: This is scary. Is `BaseModel` missing an API and therefore we need to reach into its internals?

This comment helps me a bit

```suggestion
      # AugeasTree#data -> Array
      data.data.empty?
```

imobachgs: Yes, you are right. We found that issue [some time ago](https://github.com/yast/yast-yast2/blob/ff1ba9d5371f80a6ce21f879337cd5a1f593a36f/library/general/src/lib/cfa/sysctl.rb#L121-L124) but we did not address it.

src/lib/cfa/ssg_apply.rb
imobachgs: We rarely use a Set. I am not against it, but I am not sure which are the benefits given that most of the time we use arrays.

src/lib/y2security/security_policies/manager.rb
```diff
@@ -109,6 +110,29 @@ def failing_rules(config, scope: nil, include_disabled: false)
...
 
+      # Writes custom configuration for the ssg-apply script
+      #
+      # YaST installs the package ssg-apply if a security policy is enabled. That package provides
+      # an script to be run on the first boot. The ssg-apply script invokes the oscap command using
+      # options provided in the /etc/ssg-apply/override.conf file (if it exists), or in the
+      # /etc/ssg-apply/default.conf file (if override.conf does not exist). Using an override.conf
+      # file allows for custom configuration without modifying the default configuration file.
+      #
+      # Only the #profile and #disabled-rules options are written.
```
imobachgs, line 121: I would add that, by now, we are considering only one profile.

imobachgs (commented):
LGTM

mvidner (approved):

PR #138, opened 2022-09-22
Config for ssg-apply script

joseivanlopez:
## Problem

YaST writes the configuration for the remediation script provided [*ssg-apply*]( https://drivers.suse.com/suse/SCAP/SUSE_SSG_Apply-1.0/sle-15-sp3-x86_64/install-readme.html) to the */etc/ssg-apply/override.conf* file, see  https://github.com/yast/yast-security/pull/137. YaST writes it in order to indicate to other tools what rules should be ignored. Now on *ssg-apply* will provide a new script for "disabling" rules (`mask-rule.sh`), and the configuration expected by such a script is slightly different to the config writen by YaST. 

Right now, YaST writes:

~~~
profile = disa_stig
disabled-rules = SLES-15-34512,SLES-15-45001
~~~

But the `mask-rule` script expects:

~~~
profile = disa_stig
remediation = /usr/share/scap-security-guide/bash/sle15-script-stig.sh
disabled-rules = partition_for_home,encrypted_partitions
~~~

So, *remediation* is missing and *disabled-rules* should contain rule names instead of ids.

## Solution

Adapt code to please the configuration expected by `mask-rule` script. Moreover, if no policy is enabled, then the configuration is not writen at all. 

## Testing

* Added new unit tests
* Tested manually

### DIFF
--- a/src/lib/y2security/security_policies/bootloader_password_rule.rb
+++ b/src/lib/y2security/security_policies/bootloader_password_rule.rb
@@ -39,7 +39,7 @@ module Y2Security
         end
 
         # TRANSLATORS: security policy rule
-        description = _("Bootloader must be protected by password and menu editing must be " \
+        description = _("Boot loader must be protected by password and menu editing must be " \
           "restricted")
 
         super(name, id: id, description: description, scope: :bootloader)
--- a/src/lib/y2security/security_policies/bootloader_password_rule.rb
+++ b/src/lib/y2security/security_policies/bootloader_password_rule.rb
@@ -29,15 +29,20 @@ module Y2Security
 
         # DISA STIG defines a rule for UEFI (SLES-15-010200) and another rule for non-UEFI
         # (SLES-15-010190). The condition to check both rules in YaST is exactly the same, so let's
-        # simply adapt the rule ID according to the system type.
-        id = Y2Storage::Arch.new.efiboot? ? "SLES-15-010200" : "SLES-15-010190"
+        # simply adapt the rule name and ID according to the system type.
+        if Y2Storage::Arch.new.efiboot?
+          name = "grub2_uefi_password"
+          id = "SLES-15-010200"
+        else
+          name = "grub2_password"
+          id = "SLES-15-010190"
+        end
 
-        super(
-          id,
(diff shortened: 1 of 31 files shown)


src/lib/y2security/security_policies/disa_stig_policy.rb
```diff
@@ -41,16 +41,20 @@ def initialize
...
-          MissingMountPointRule.new("SLES-15-040200", "/home"),
-          MissingMountPointRule.new("SLES-15-040210", "/var"),
-          SeparateFilesystemRule.new("SLES-15-030810", "/var/log/audit"),
-          FilesystemSizeRule.new("SLES-15-030660", "/var/log/audit",
-            min_size: Y2Storage::DiskSize.MiB(100)),
+          MissingMountPointRule.new("partition_for_home", "SLES-15-040200", "/home"),
+          MissingMountPointRule.new("partition_for_var", "SLES-15-040210", "/var"),
+          SeparateFilesystemRule.new("partition_for_var_log_audit",
+            "SLES-15-030810", "/var/log/audit"),
+          FilesystemSizeRule.new("auditd_audispd_configure_sufficiently_large_partition",
```
imobachgs, line 56: Oh, a rather small name :-)

src/lib/y2security/security_policies/bootloader_password_rule.rb
```diff
@@ -29,15 +29,20 @@ def initialize
...
+        if Y2Storage::Arch.new.efiboot?
+          name = "grub2_uefi_password"
+          id = "SLES-15-010200"
+        else
+          name = "grub2_password"
+          id = "SLES-15-010190"
+        end
+
+        # TRANSLATORS: security policy rule
+        description = _("Bootloader must be protected by password and menu editing must be " \
```
imobachgs, line 42: np: I would use "boot loader" (I have checked that we use that form rather often in our UI).

imobachgs (approved):
Just a small comment about how to write "boot loader".

imobachgs (approved):

imobachgs (approved):
LGTM. Thanks!

PR #142, opened 2022-10-04
Set the action to be performed by SCAP on first boot

imobachgs:
## Summary

This PR adds support for setting the action to be performed using SCAP during the first boot, among other changes.

## Changes

* Allow setting an action for the first boot (*nothing*, *scan only* or *scan and remediate*).
* Report an error instead of blocking the installation when the security policy is not met.
* Changes in `ssg-apply` configuration:
  * Update the `remediate` value (`yes` for `scan and remediate`; `no` for `scan` only).
  * Do not write the remediation script and the list of disabled rules to the `ssg-apply` configuration.
  * Use `/etc/ssg-apply/default.conf` as base for the configuration file.
* Write the list of failing rules to `/var/log/YaST2/security_policies_failed_rules`.
* Add a help text specifying how the feature works.

## Screenshots

<details>
<summary>Summary including the new feature to set the SCAP action on first boot</summary>

stig-scap-action
</details>

<details>
<summary>Help text explaining how the feature works</summary>

stig-help-text

</details>

<details>
<summary>Warn the user but do not block the installation</summary>

stig-error-stop
</details>

<details>
<summary>List of failed rules</summary>

stig-failed-rules
</details>

### DIFF
--- a/src/lib/y2security/clients/security_policy_proposal.rb
+++ b/src/lib/y2security/clients/security_policy_proposal.rb
@@ -154,11 +154,11 @@ module Y2Security
 
       # Returns the warning level
       #
-      # Blocker if there are enabled failing rules
+      # Error if there are enabled failing rules
       #
-      # @return [Symbol] :blocker
+      # @return [Symbol,nil] :error or nil
       def warning_level
-        success? ? nil : :blocker
+        success? ? nil : :error
       end
 
       # Runs the security policies checks
--- a/src/lib/y2security/clients/security_policy_proposal.rb
+++ b/src/lib/y2security/clients/security_policy_proposal.rb
@@ -337,7 +337,7 @@ module Y2Security
         #
         # @return [Array<String>]
         def links_for_scap_actions
-          Y2Security::SecurityPolicies::Manager.scap_actions.map do |action|
+          Y2Security::SecurityPolicies::Manager.known_scap_actions.map do |action|
             scap_action_link(action)
           end
         end
@@ -421,14 +421,13 @@ module Y2Security
         def to_html
           toggle_link = links_builder.policy_toggle_link(policy)
 
-          sections = []
-          sections << if policy_enabled
+          if policy_enabled
             format(
               # TRANSLATORS: 'policy' is a security policy name; 'link' is just an HTML-like link
               _("%{policy} is enabled (<a href=\"%{link}\">disable</a>)"),
               policy: policy.name,
               link:   toggle_link
-            )
+            ) + scap_action_description + rules_section
           else
             format(
               # TRANSLATORS: 'policy' is a security policy name; 'link' is just an HTML-like link
@@ -437,12 +436,6 @@ module Y2Security
               link:   toggle_link
             )
           end
-
-          if policy_enabled
-            sections << scap_action_description
-            sections << rules_section
-          end
-          sections.join
         end
 
       private
--- a/src/lib/y2security/clients/security_policy_proposal.rb
+++ b/src/lib/y2security/clients/security_policy_proposal.rb
@@ -90,6 +90,8 @@ module Y2Security
           toggle_rule(id)
         when "fix-rule"
           fix_rule(id)
+        when "set-scap-action"
+          policies_manager.scap_action = id.to_sym
         when "storage"
           result = open_client("inst_disk_proposal")
         when "bootloader"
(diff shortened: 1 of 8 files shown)


src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -309,6 +333,15 @@ def links_for_rule(rule)
           ]
         end
 
+        # Possible links to set the SCAP action
+        #
+        # @return [Array<String>]
+        def links_for_scap_actions
```
joseivanlopez, line 339: NP: I would try to keep the same name style as the rest of methods, e.g., `#scap_actions_links`.

src/lib/y2security/clients/security_policy_proposal.rb
```diff
@@ -393,6 +437,12 @@ def to_html
               link:   toggle_link
             )
           end
+
+          if policy_enabled
```
joseivanlopez, line 441: NP: Why cannot we add these sections in the if above?

src/lib/y2security/security_policies/manager.rb
```diff
@@ -21,19 +21,35 @@
...
 Yast.import "PackagesProposal"
-
-Yast.import "PackagesProposal"
+Yast.import "Service"
 
 module Y2Security
   module SecurityPolicies
     # Class to manage security policies
     class Manager
+      # @return [Symbol] action to perform after the installation. :remediate peforms a full
```
joseivanlopez, line 33: It can be document with exact values: `@return [:none, :scan, :remediate]`.

src/lib/y2security/security_policies/manager.rb
```diff
@@ -120,21 +159,39 @@ def failing_rules(config, scope: nil, include_disabled: false)
...
-        policy = policies.find { |p| enabled_policy?(p) }
-
-        return unless policy
-
+      # Bear in mind that ssg-apply does not perform any kind of merging between both configuration
+      # files.
+      #
+      # Only the #profile and #remediate options are written.
+      def write_config(policy)
+        copy_default_config
```
joseivanlopez, line 167: NP: could we have something like `file.save_as`? This would allow reading from one path, edit, and then save as another file.

src/lib/y2security/security_policies/manager.rb
```diff
@@ -106,12 +130,27 @@ def enabled_policy?(policy)
       #   only display rules that it can fix.
       # @return [Hash<Policy, Array<Rule>>]
       def failing_rules(config, scope: nil, include_disabled: false)
-        enabled_policies.each_with_object({}) do |policy, result|
+        @last_result = enabled_policies.each_with_object({}) do |policy, result|
```
joseivanlopez, line 133: Do we need to catch this? Checking rules is quite fast and there is some chance to use outdated results.

joseivanlopez (commented):
Looks good! Only some details and questions.

src/lib/y2security/security_policies/manager.rb
imobachgs: We are catching the results because we need them *at the end* of the installation. But if the TargetConfig remains the same, we can avoid it.

src/lib/y2security/clients/security_policy_proposal.rb
imobachgs: I thought the style was `links_*`. We have `links_for_rule` and `links_for_policy` already. See https://github.com/yast/yast-security/blob/190360d82b2d8b5efa24cd42bf139e7afe510abe/src/lib/y2security/clients/security_policy_proposal.rb#L132-L134 for its usage.

src/lib/y2security/clients/security_policy_proposal.rb
imobachgs: Yes, you are right. I changed this code a few times, but the final result does not make sense :-) Fixed.

src/lib/y2security/security_policies/manager.rb
imobachgs: Yes, that was my first thought. But we need to fiddle with the internals of the `BaseModel` class (a rather fragile approach) or change that class. And, to be honest, I would rather prefer to limit these changes.

joseivanlopez (approved):
LGTM

PR #144, opened 2022-10-18
Use "stig" as profile name for DISA STIG

imobachgs:
## Problem

The profile "disa_stig" does not exists.

## Solution

Use `stig` instead of `disa_stig` as profile name for ssg-apply configuration.


joseivanlopez (approved):
LGTM

PR #147, opened 2022-11-08
Disable ssg-apply service if SCAP action is :none

imobachgs:
## Problem

`ssg-apply` even if the user asked to do nothing during the 1st boot. The problem is `ssg-apply` enables itself using a post-script (see https://build.suse.de/package/view_file/home:andavis:GEHC:stig/ssg-apply/ssg-apply.spec?expand=1) and YaST does not disable the service (it does not expect that behaviour).

## Solution

Explicitly disable the `ssg-apply`. Morever, the package should be adjusted to not enable itself.


joseivanlopez (approved):
LGTM

PR #151, opened 2022-12-09
Export security policy settings

imobachgs:
## Problem

The `security_policy` section is not exported when cloning a system with AutoYaST.

## Solution

If the `/etc/ssg-apply/override.conf` file exists (written by AutoYaST), infer and export the settings.

* If the `ssg-apply` is disabled, set the `action` to "none".
* If the `ssg-apply` is enabled...
  * ... and `remediate` is set to `yes` in the configuration, set the action to `remediate`.
  * ... and `remediate` is set to `no` in the configuration, set the action to `scan`.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/src/lib/y2security/autoinst_profile/security_policy_section.rb
+++ b/src/lib/y2security/autoinst_profile/security_policy_section.rb
@@ -18,6 +18,10 @@
 # find current contact information at www.suse.com.
 
 require "installation/autoinst_profile/section_with_attributes"
+require "y2security/security_policies/manager"
+require "cfa/ssg_apply"
+
+Yast.import "Service"
 
 module Y2Security
   module AutoinstProfile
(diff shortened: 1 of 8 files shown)


src/lib/y2security/autoinst_profile/security_policy_section.rb
```diff
@@ -42,6 +46,21 @@ def self.attributes
...
       #   @return [String,nil] Enabled policy
+
+      # Clones the security policy settings from the underlying system
+      def self.new_from_system
+        file = CFA::SsgApply.load
+        section = new
+        return section if file.empty?
+
+        section.action = if Y2Security::SecurityPolicies::Manager.instance.service_enabled?
+          (file.remediate == "yes") ? "remediate" : "scan"
```
jreidinger, line 57: I would prefer here if it can be value per condition as it is more readable so something like:
```
section.action = if !Y2Security::SecurityPolicies::Manager.instance.service_enabled?
  "none"
elsif file.remediate == "yes"
  "remediate"
else
  "scan"
end
```
this way I found it more readable.

jreidinger (approved):
just NP

jreidinger (approved):

PR #152, opened 2022-12-20
Always write override conf

imobachgs:
## Problem

AutoYaST does not clone the `<security_policy/>` section when the action is set to `none`:

```xml
<security_policy>
  <action>none</action>
  <profile>stig</profile>
</security_policy>
```

In that case, YaST does not write an `/etc/ssg-apply/override.conf` file, so there is no way to infer that security policy checks were enabled at installation time. Checking whether `ssg-apply` package is installed does not look like a valid way to determine it because you might get a `<security_policy/>` section just for installing the package.

## Solution

Starting in `yast2-security 4.4.19`, if a security policy is enabled at installation time, YaST *always* writes the `override.conf` file. So when cloning the system, it exports a `<security_policy/>` section if that file exists.

## Towards a better solution

This solution is good enough, but we would rather prefer to rely on a YaST-specific file. In that regard, we suggest using a `/etc/ssg-apply.d` directory where we could drop a YaST-specific file (e.g., `70-yast.conf`). That file would tell us that security policies were enabled during installation. Moreover, `ssg-apply` would behave similarly to other tools (`ls -d /etc/*.d /usr/etc/*.d`).

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Dec 20 10:21:49 UTC 2022 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Always write the ssg-apply configuration if a security policy
+  is enabled, even if the action is 'none' (related to
+  jsc#SLE-24764).
+- 4.4.19
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


joseivanlopez (approved):
LGTM

PR #153, opened 2022-12-20
Merge SLE-15-SP4

imobachgs:
Bring #151 and #152 to `master`.

### DIFF
--- a/test/y2security/autoinst_profile/security_policy_section_test.rb
+++ b/test/y2security/autoinst_profile/security_policy_section_test.rb
@@ -41,4 +41,55 @@ describe Y2Security::AutoinstProfile::SecurityPolicySection do
       end
     end
   end
+
+  describe ".new_from_system" do
+    let(:service_enabled?) { true }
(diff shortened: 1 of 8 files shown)


test/y2security/autoinst_profile/security_policy_section_test.rb
```diff
@@ -41,4 +41,55 @@
...
+    context "when the ssg-apply service does not exist" do
+      let(:file) { instance_double(CFA::SsgApply, empty?: true) }
+
+      it "returns an empty section" do
+        section = described_class.new_from_system
+        expect(section.to_hashes).to be_empty
+      end
+    end
+
+    context "when the ssg-apply service exists is disabled" do
```
teclator, line 62: NP: Would suggest something like, but as this is a merge PR just merge it as it is...

```suggestion
    context "when the ssg-apply service exists but it is disabled" do
```

teclator (approved):
LGTM

PR #162, opened 2026-02-05
write proper /etc/selinux/config, according to expected format

rfrohl:
Proper format should not carry additional spaces and should look like 

```
$ cat /etc/selinux/config
SELINUX=disabled
```

## Problem

The config file format that is generated will not be parsed by the selinux toolchain

- https://bugzilla.suse.com/show_bug.cgi?id=1257752

## Solution

Removes unnecessary whitespaces.

## Testing

- *Tested manually*

## Screenshots

*If the fix affects the UI attach some screenshots here.*

### DIFF
--- a/package/yast2-security.changes
+++ b/package/yast2-security.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Feb  5 13:18:03 UTC 2026 - Robert Frohl <rfrohl@suse.com>
+
+- write proper /etc/selinux/config according to expected format
+  (bsc#1257752)
+- 5.0.4
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


package/yast2-security.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Feb  5 13:18:03 UTC 2026 - Robert Frohl <rfrohl@suse.com>
+
+- write proper /etc/selinux/config according to expected format 
+  (bsc#1257752)
```
jreidinger: please also bump version like other changes. This ensure that after merge it will be send to factory. ( change itself looks good )

jreidinger (approved):

PR #163, opened 2026-02-06
fix changes format

jreidinger:
## Problem

autosubmission does not work due to format issue in changelog that is now forced to be compliant.

See https://github.com/yast/yast-security/pull/162


teclator (approved):
LGTM

teclator (approved):
LGTM

